PR #5, opened 2023-03-23
cmake: Modernize the CMake of the utility library

juan-lunarg:


CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+endif()
+
+find_package(PythonInterp 3.7.2 REQUIRED)
+
+add_subdirectory(scripts)
+
+find_package(VulkanHeaders REQUIRED CONFIG QUIET)
+
+add_library(VulkanLayerUtils STATIC)
+add_library(Vulkan::LayerUtils ALIAS VulkanLayerUtils)
```
christophe-lunarg, line 47: These specifics approach is quite annoying for a Visual Studio developer because Visual Studio use the CMakeLists.txt where this is defined instead of the one in `src` where all the configuration for these targets are done.

juan-lunarg: You can add multiple CMakeLists.txt with target_sources and then use `source_groups` to fix the IDE presentation.

PR #7, opened 2023-03-27
Add GitHub actions

christophe-lunarg:
I fixed (trivial) the case for Windows tests from https://github.com/KhronosGroup/Vulkan-Utility-Libraries/pull/5

### DIFF
--- a/.github/workflows/vul.yml
+++ b/.github/workflows/vul.yml
@@ -1,0 +1,43 @@
+# Copyright (c) 2023 Valve Corporation
+# Copyright (c) 2023 LunarG, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


.github/workflows/vul.yml
```diff
@@ -0,0 +1,43 @@
...
+      matrix:
+        config: [Debug, Release]
+        os: [ ubuntu-latest, windows-latest, macos-latest ]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.8'
+      - name: Configure
+        run: cmake -S. -B build -D VUL_WERROR=ON -D VUL_BUILD_TESTS=ON -D CMAKE_BUILD_TYPE=${{matrix.config}} -D UPDATE_DEPS=ON
```
juan-lunarg, line 38: `VUL_WERROR`/`VUL_BUILD_TESTS` doesn't exist yet.

juan-lunarg (changes requested):

.github/workflows/vul.yml
christophe-lunarg: This is fixed with last commit

juan-lunarg (approved):

PR #9, opened 2023-03-29
cmake: Use VUL_ prefix for cache variables

juan-lunarg:
Allows build options specific to Vulkan-Utility-Libraries to be easily grouped via cmake-gui.

This approach is also nicer for users who consume the library via add_subdirectory or FetchContent since it more obvious what settings are being ovverriden.

Remove UTILITY_LIBRARIES_CPP_STANDARD since we don't have a reason to support multiple CPP standards.

Set testing to OFF by default to make consuming the library easier for users and package managers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,19 +12,17 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 cmake_minimum_required(VERSION 3.17.2)
 
-project(VULKAN_UTILITY_LIBRARIES LANGUAGES CXX)
+project(VUL LANGUAGES CXX)
 
 include(GNUInstallDirs)
 
 set_property(GLOBAL PROPERTY USE_FOLDERS ON) # Remove when min is 3.26, see CMP0143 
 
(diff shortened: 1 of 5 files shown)


juan-lunarg:
image

Note how using a common prefix nicely groups things together in cmake-gui which is a nice quality of life feature for users unfamiliar with the repository.

NOTE: `UPDATE_DEPS` isn't prefixed with `VUL_` to avoid confusion with the other Khronos/LunarG repos. Having commonality matters more in that instance.

CMakeLists.txt
```diff
@@ -12,19 +12,17 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 cmake_minimum_required(VERSION 3.17.2)
 
-project(VULKAN_UTILITY_LIBRARIES LANGUAGES CXX)
+project(VUL LANGUAGES CXX)
```
juan-lunarg, line 17: NOTE: Changing the project name to be brief also makes accessing the variables generated by the project command simpler.

EX:

`VUL_SOURCE_DIR` vs `VULKAN_UTILITY_LIBRARIES_SOURCE_DIR`

christophe-lunarg: ok

christophe-lunarg (approved):

PR #14, opened 2023-04-20
Implement layer settings library

christophe-lunarg:
The goal of this PR is to implement the layer settings library that will be used by all layers.

This includes and implementation for `VK_EXT_layer_settings` to be able to set layer settings programmatically.


include/CMakeLists.txt
```diff
@@ -12,8 +12,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-target_include_directories(VulkanLayerUtils PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>)
+target_include_directories(VulkanLayerSettings PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>)
 
-target_compile_features(VulkanLayerUtils PUBLIC cxx_std_17)
+target_compile_features(VulkanLayerSettings PUBLIC cxx_std_17)
```
juan-lunarg, line 17: Looking at vk_layer_settings.h this can be removed now. Since vk_layer_setting.h exposes a C interface now.

Instead it should now be `target_compile_features(VulkanLayerSettings PRIVATE cxx_std_17)` and moved to the `src` directory since it's an implementation detail.

include/vulkan/layer/vk_layer_settings.h
```diff
@@ -21,44 +21,178 @@
...
+
+#define VK_EXT_layer_settings 1
+#define VK_EXT_LAYER_SETTINGS_SPEC_VERSION 1
+#define VK_EXT_LAYER_SETTINGS_EXTENSION_NAME "VK_EXT_layer_settings"
+
+// This extension is exclusively used by VVL, and is NOT intended as a deliverable.
+// The value of the VK_STRUCTURE_TYPE is arbitrary. The only requirement,
+// is that it must not conflict with existing sTypes.
+//
+// NOTE: VK_STRUCTURE_TYPE_MAX_ENUM - 1 is used by the intel driver.
```
juan-lunarg, line 85: \+ @mark-lunarg since I don't remember the full context.

We would eventually want to make an official extension to get a proper sType. Right?

include/vulkan/layer/vk_layer_settings.h
```diff
@@ -21,44 +21,178 @@
...
+constexpr VkStructureType VK_STRUCTURE_TYPE_INSTANCE_LAYER_SETTINGS_EXT =
```
juan-lunarg, line 86: constexpr isn't C friendly

include/vulkan/layer/vk_layer_settings.h
```diff
@@ -21,44 +21,178 @@
...
+typedef struct VkLayerSettingsCreateInfoEXT {
+    VkStructureType sType;
+    void *pNext;
+    uint32_t settingCount;
+    VkLayerSettingEXT *pSettings;
+} VkLayerSettingsCreateInfoEXT;
 
-// Query setting data for FRAMES setting type in the layer manifest
-std::string GetLayerSettingFrames(const char *layer_key, const char *setting_key);
+//typedef std::vector<std::pair<std::string, int>> List;
```
juan-lunarg, line 187:
```suggestion
```

tests/find_package/CMakeLists.txt
```diff
@@ -13,8 +13,8 @@ find_package(VulkanHeaders REQUIRED CONFIG)
...
 find_package(VulkanUtilityLibraries REQUIRED CONFIG)
 
-if (NOT TARGET Vulkan::LayerUtils)
-    message(FATAL_ERROR "Vulkan::LayerUtils target not defined!")
+if (NOT TARGET Vulkan::LayerSettings)
+    message(FATAL_ERROR "Vulkan::LayerSettings target not defined!")
 endif()
 
-target_link_libraries(find_package_example PRIVATE Vulkan::LayerUtils)
+target_link_libraries(find_package_example PRIVATE Vulkan::LayerSettings)
```
juan-lunarg, line 20: We should make find_package_example a C project to ensure the interface is C compatible. We just need to change the languages to C, and `client.cpp` to `client.c`

We can keep `add_subdirectory_example` as is to ensure the header file compiles with C++ correctly as well.

` Vulkan is a C99 API designed for explicit control of low-level graphics and compute functionality.`

src/layer/vk_layer_settings.cpp
```diff
@@ -0,0 +1,203 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Authors:
+ * - Christophe Riccio <christophe@lunarg.com>
+ * - Mark Lobodzinski <mark@lunarg.com>
+ * - Jon Ashburn
+ * - Courtney Goeltzenleuchter
+ * - Tobin Ehlis
```
juan-lunarg, line 23: Only you and @mark-lunarg seem applicable.

include/vulkan/layer/vk_layer_settings.h
```diff
@@ -21,44 +21,178 @@
...
 #pragma once
 
-#include <string>
-#include <vector>
-
 #include <vulkan/vk_layer.h>
 #include <vulkan/vulkan.h>
 
-namespace vku {
+// VK_EXT_layer_settings
```
juan-lunarg, line 27: Ensure the interface is C friendly.

```suggestion
#ifdef __cplusplus
extern "C" {


...


#ifdef __cplusplus
}
#endif
// VK_EXT_layer_settings
```

include/vulkan/layer/vk_layer_settings.h
```diff
@@ -21,44 +21,178 @@
 
 #pragma once
 
-#include <string>
-#include <vector>
-
 #include <vulkan/vk_layer.h>
 #include <vulkan/vulkan.h>
```
juan-lunarg, line 25: Do we need to expose both? I think just `vk_layer.h` is enough for the interface.

tests/add_subdirectory/client.cpp
```diff
@@ -1,3 +1,3 @@
 #include <vulkan/layer/vk_layer_settings.h>
 
-bool foobar() { return vku::IsLayerSetting("layer_name", "setting_key"); }
\ No newline at end of file
+bool foobar() { return vlHasLayerSetting("setting_key"); }
```
juan-lunarg, line 3: This is bike shedding but should the prefix be `vl`?

Just as other options.

`vul` for `vulkan utility libraries`

`vulkanLayerSettings` for verbosity.

~~vk~~ I think should be off the table since it's not part of the official vulkan headers.

vl also makes sense for `vulkan layer` though. So I don't really have a strong opinion to be honest.

tests/find_package/CMakeLists.txt
juan-lunarg:
```
cmake_minimum_required(VERSION 3.17)

project(TEST_FIND_PACKAGE LANGUAGES C)

add_library(find_package_example STATIC)

# Test c99 support as well as find_package support
target_compile_features(find_package_example PRIVATE c_std_99)

target_sources(find_package_example PRIVATE
    ${CMAKE_CURRENT_LIST_DIR}/../add_subdirectory/client.c
)

# NOTE: Because VulkanHeaders is a PUBLIC dependency it needs to be found prior to VulkanUtilityLibraries
find_package(VulkanHeaders REQUIRED CONFIG)

find_package(VulkanUtilityLibraries REQUIRED CONFIG)

if (NOT TARGET Vulkan::LayerSettings)
    message(FATAL_ERROR "Vulkan::LayerSettings target not defined!")
endif()

target_link_libraries(find_package_example PRIVATE Vulkan::LayerSettings)
```

include/vulkan/layer/vk_layer_settings.h
christophe-lunarg: Yes this extension will become official but first we decided to implement it and tests it, potentially iterate on its definition.

tests/find_package/CMakeLists.txt
christophe-lunarg: Great idea!

include/CMakeLists.txt
christophe-lunarg: This is resolved with last commit

include/vulkan/layer/vk_layer_settings.h
christophe-lunarg: Fixed

include/vulkan/layer/vk_layer_settings.h
christophe-lunarg: Fixed

include/vulkan/layer/vk_layer_settings.h
christophe-lunarg: Fixed

tests/find_package/CMakeLists.txt
christophe-lunarg: Done

tests/add_subdirectory/client.cpp
christophe-lunarg: That remains an open question to me too! And I was wondering what you where planning for the validation layer libraries so we could create a consistent approach for all this.

src/layer/vk_layer_settings.cpp
christophe-lunarg: Fixed

tests/find_package/CMakeLists.txt
```diff
@@ -1,20 +1,38 @@
...
+# Test c99 support as well as find_package support
+target_compile_features(find_package_example PRIVATE c_std_99)
+
 target_sources(find_package_example PRIVATE
-    ${CMAKE_CURRENT_LIST_DIR}/../add_subdirectory/client.cpp
+    ${CMAKE_CURRENT_LIST_DIR}/../add_subdirectory/client.c
 )
 
-# NOTE: Because VulkanHeaders is a PUBLIC dependency it needs to be found prior to VulkanUtilityLibraries
+# NOTE: Because VulkanHeaders is a PUBLIC dependency it needs to be found prior to VulkanLayerSettings
```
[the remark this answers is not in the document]
christophe-lunarg: Fixed

tests/add_subdirectory/client.cpp
christophe-lunarg: So using a single namespace for everything?

juan-lunarg: I don't have too much frame of reference for a package that exports multiple libraries. I've never really dealt with Boost or Qt so I don't know what the expected standard is.

juan-lunarg: I think for now the namespace should just be `Vl`. When another library gets added we will see.

include/vulkan/layer/vk_layer_settings.h
```diff
@@ -21,44 +21,186 @@
...
+    VK_LAYER_SETTING_TYPE_FRAMESET_EXT,
+    VK_LAYER_SETTING_TYPE_FRAMESET_ARRAY_EXT,
+} VkLayerSettingTypeEXT;
 
-// Query setting data for BOOL setting type in the layer manifest
-bool GetLayerSettingBool(const char *layer_key, const char *setting_key);
+typedef struct array_int32 {
+    uint32_t count;
+    int32_t *data;
+} array_int32;
```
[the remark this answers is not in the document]
christophe-lunarg: Yes this part is still WIP

tests/add_subdirectory/client.cpp
christophe-lunarg: Agreed!

include/vulkan/layer/vk_layer_settings.h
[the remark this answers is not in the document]
christophe-lunarg: And actually if you have better idea than this to pass array of values and strings for example, I'll take it, I am not convinved!

juan-lunarg (approved):

include/vulkan/layer/vk_layer_settings.h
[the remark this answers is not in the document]
christophe-lunarg: Ok that works with me!

include/vulkan/layer/vk_layer_settings.h
```diff
@@ -21,44 +21,24 @@
...
 #pragma once
 
-#include <string>
-#include <vector>
+#ifdef __cplusplus
+extern "C" {
+#endif 
 
+#include "vk_layer_settings_ext.h"
 #include <vulkan/vk_layer.h>
```
juan-lunarg, line 29: This include shouldn't be needed.

```suggestion
```

include/vulkan/layer/vk_layer_settings.h
```diff
@@ -21,44 +21,23 @@
...
-// Initialize the callback function to get error messages. By default the error messages are outputed to stderr. Use nullptr to
-// return to the default behavior.
-void InitLayerSettingsLogCallback(LAYER_SETTING_LOG_CALLBACK callback);
+// Check whether a setting was set either programmatically, from vk_layer_settings.txt or an environment variable
+VkBool32 vlHasLayerSetting(const char *pSettingName);
 
-// Check whether a setting was set either from vk_layer_settings.txt or an environment variable
-bool IsLayerSetting(const char *layer_key, const char *setting_key);
+// Query setting data for a setting
+void vlGetLayerSettingData(const char *pSettingName, VkLayerSettingTypeEXT type, VkLayerSettingDataEXT *pData);
```
juan-lunarg, line 39: Perhaps this would be a good approach to avoid having array types?

```suggestion
// Query setting data for a setting.
//
// Query once to get the number of elements.
//
// Allocate memory for the setting.
//
//Then call the function again to fill out the array of VkLayerSettingDataEXT.
void vlGetLayerSettingData(const char *pSettingName, VkLayerSettingTypeEXT type, uint32_t *pSettingCount, VkLayerSettingDataEXT *pSetting);
```

It would be similar in nature to `vkEnumeratePhysicalDevices` where you call once to get the number of devices, and then fill out the devices.

include/vulkan/layer/vk_layer_settings.h
```diff
@@ -21,44 +21,23 @@
...
+#include "vk_layer_settings_ext.h"
 
-namespace vku {
+typedef void *(*VL_LAYER_SETTING_LOG_CALLBACK)(const char *pSettingName, const char *pMessage);
 
-typedef std::vector<std::string> Strings;
-typedef std::vector<std::pair<std::string, int>> List;
-typedef void *(*LAYER_SETTING_LOG_CALLBACK)(const char *setting_key, const char *message);
+// Initialize the layer settings. If 'pCallback' is set to NULL, the messages are outputed to stderr.
+void vlInitLayerSettings(const char *pLayerName, const void *pNext, VL_LAYER_SETTING_LOG_CALLBACK pCallback);
```
juan-lunarg, line 33: Instead of returning void couldn't this return VK_RESULT.

If given an invalid layer perhaps returning `VK_ERROR_LAYER_NOT_PRESENT` would make sense.

include/vulkan/layer/vk_layer_settings.h
christophe-lunarg: Interresting that you are suggesting this approach because this is exactly the declaration I was working from on Friday because the array approach doesn't allow me implementing the array types.

include/vulkan/layer/vk_layer_settings.h
christophe-lunarg: I am also using VK_RESULT for `vlGetLayerSettingData` current investigation but yes that sounds like a good idea for `vlInitLayerSettings` too

PR #16, opened 2023-05-08
known_good: Update Vulkan Headers to 1.3.250

juan-lunarg:


juan-lunarg:
'Needed' for https://github.com/KhronosGroup/Vulkan-Profiles/pull/419

Not sure if this is actually needed. However, for our SDK releases we have scripts that produce warnings if the version of vulkan headers in known_good.json isn't consistent.

christophe-lunarg (approved):

PR #20, opened 2023-05-23
layer: fix const to VkLayerSettingsCreateInfoEXT

christophe-lunarg:
Without this fix, the profiles layer can't set the settings properly...


juan-lunarg (approved):
LGTM but why does this fix things?

PR #22, opened 2023-05-24
layer: avoid conflict with validation layer setting extension

christophe-lunarg:
The validation layer is already using VK_STRUCTURE_TYPE_INSTANCE_LAYER_SETTINGS_EXT with a specific value. We need a different value and name to avoid the layer setting library to interpret some structure which are specificed differently causing a segfault.

### DIFF
--- a/include/vulkan/layer/vk_layer_settings_ext.h
+++ b/include/vulkan/layer/vk_layer_settings_ext.h
@@ -90,7 +90,7 @@ extern "C" {
 // is that it must not conflict with existing sTypes.
 //
 // NOTE: VK_STRUCTURE_TYPE_MAX_ENUM - 1 is used by the intel driver.
-#define VK_STRUCTURE_TYPE_INSTANCE_LAYER_SETTINGS_EXT ((VkStructureType)(VK_STRUCTURE_TYPE_MAX_ENUM - 42))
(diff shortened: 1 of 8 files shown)


include/vulkan/layer/vk_layer_settings_ext.h
```diff
@@ -90,7 +90,7 @@ extern "C" {
 // is that it must not conflict with existing sTypes.
 //
 // NOTE: VK_STRUCTURE_TYPE_MAX_ENUM - 1 is used by the intel driver.
-#define VK_STRUCTURE_TYPE_INSTANCE_LAYER_SETTINGS_EXT ((VkStructureType)(VK_STRUCTURE_TYPE_MAX_ENUM - 42))
+#define VK_STRUCTURE_TYPE_LAYER_SETTINGS_EXT ((VkStructureType)(VK_STRUCTURE_TYPE_MAX_ENUM - 43))
```
juan-lunarg, line 93:
```suggestion
// NOTE: VK_STRUCTURE_TYPE_MAX_ENUM - 42 is used by the validation layers
#define VK_STRUCTURE_TYPE_LAYER_SETTINGS_EXT ((VkStructureType)(VK_STRUCTURE_TYPE_MAX_ENUM - 43))
```

juan-lunarg (commented):
Add a comment to avoid the conflict.

juan-lunarg (approved):

PR #31, opened 2023-06-20
cmake: Modernize update_deps.py logic

juan-lunarg:
Pulling in latest changes from VVL


CMakeLists.txt
```diff
@@ -45,8 +45,8 @@ add_subdirectory(src)
 add_subdirectory(include)
 
 if (VUL_IS_TOP_LEVEL)
-    option(VUL_TESTS "Build tests")
-    if (VUL_TESTS)
+    option(BUILD_TESTS "Build tests")
```
juan-lunarg, line 48: Note: Since I continually am updating the update_deps.py logic using the same BUILD_TESTS variable across repos helps quite a bit. Also because this option is only exposed if the project is top level it doesn't negatively impact `add_subdirectory` users.

scripts/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
+# ~~~
```
juan-lunarg, line 1: Changes to this file improve performance quite a bit and uses the modern approach for finding python. It also has much better android support.

PR #32, opened 2023-06-20
ci: Add android to CI

juan-lunarg:


src/layer/layer_settings_manager.cpp
```diff
@@ -51,6 +51,8 @@ static std::string GetAndroidProperty(const char *name) {
         __system_property_read_callback(
             pi,
             [](void *cookie, const char *name, const char *value, uint32_t serial) {
+                (void)name;
+                (void)serial;
```
juan-lunarg, line 55: clang complained about unused variables.

Issue #13, opened 2023-04-07
Remove VLL as a dependency of Vulkan-Profiles


juan-lunarg:
Context:
https://github.com/KhronosGroup/Vulkan-Profiles/pull/399#issuecomment-1498004870

Note in order to fully remove the dependency on the validation layers it looks like we need to add functionality that covers:

- [x] vk_layer_dispatch_table.h ( generated)
- [x] vk_dispatch_table_helper.h (generated)
- [x] vk_layer_utils.h (only a small amount of the file is actually needed)
- [x] vk_enum_string_helper.h (generated)

juan-lunarg:
Minor update.

Looked into this last week. The main difficulty is wrapping my head around the code generation in the vulkan validation layers. It's complex, involved, and a bit inconsistent.

Also the code generation in vvl assumes all the generated code will be placed in one location (`layers/generated`). Which practically speaking cannot be the case for the utility libraries.

juan-lunarg:
Note. We don't necessarily want 1 to 1 parity of what's in VVL right now.

We need to understand how we can make the logic being used from VVL generic / user friendly.

juan-lunarg:
MVP @charles-lunarg

PR #41, opened 2023-07-13
scripts: Add initial codegen scripts

spencer-lunarg:
For https://github.com/KhronosGroup/Vulkan-Utility-Libraries/issues/13

I didn't run and check in the code as need to change where they will go

`vulkan_object.py` is the same as VVL, `base_generator.py` was improved and going to port changes back


scripts/generate_source.py
```diff
@@ -0,0 +1,110 @@
...
+    from generators.base_generator import BaseGeneratorOptions
+    from generators.dispatch_table_helper_generator import DispatchTableHelperOutputGenerator
+    from generators.layer_dispatch_table_generator import LayerDispatchTableOutputGenerator
+    from generators.enum_string_helper_generator import EnumStringHelperOutputGenerator
+
+    # Build up a list of all generators and custom options
+    generators = {
+        'vk_dispatch_table_helper.h' : {
+            'generator' : DispatchTableHelperOutputGenerator,
+            'directory' : 'include/vulkan/layer',
```
spencer-lunarg, line 42: @juan-lunarg this is where we can simply change the location

juan-lunarg (approved):

scripts/generate_source.py
juan-lunarg: Gotcha 👍🏾

spencer-lunarg: @juan-lunarg  Do you know where you want these, then we can set them, run the script and check in the generated code

juan-lunarg: Since these are all header only they should go in the `include/vulkan` directory.

juan-lunarg: FYI @mikes-lunarg

PR #45, opened 2023-07-21
build: Remove invalid generated files

juan-lunarg:
vk_dispatch_table_helper.h and vk_layer_dispatch_table.h don't compile currently.

Added minimal compilation testing for vk_enum_string_helper.h


include/vulkan/vk_dispatch_table_helper.h
```diff
@@ -1,2354 +0,0 @@
...
-****************************************************************************/
-// NOLINTBEGIN
-#pragma once
-
-#include <vulkan/vulkan.h>
-#include <vulkan/vk_layer.h>
-#include <cstring>
-#include <string>
-#include "vk_layer_dispatch_table.h"
-#include "vk_extension_helper.h"
```
juan-lunarg, line 30: vk_extension_helper.h refers to a file that only exists in vvl which in turn refers to even more custom containers

include/vulkan/vk_layer_dispatch_table.h
```diff
@@ -1,783 +0,0 @@
...
-* Unless required by applicable law or agreed to in writing, software
-* distributed under the License is distributed on an "AS IS" BASIS,
-* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-* See the License for the specific language governing permissions and
-* limitations under the License.
-****************************************************************************/
-// NOLINTBEGIN
-#pragma once
-
-typedef PFN_vkVoidFunction (VKAPI_PTR *PFN_GetPhysicalDeviceProcAddr)(VkInstance instance, const char* pName);
```
juan-lunarg, line 25: This file also fails to compile on it's own. It require the user provide header before hand which isn't ideal for users.

PR #58, opened 2023-08-08
Add GN build system

christophe-lunarg:
I have no idea what I am doing here or what I am supose to do let alone where to find information on GN...

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -1,0 +1,36 @@
+# Copyright 2023-2023 The ANGLE Project Authors.
+# Copyright 2023-2023 LunarG, Inc.
+#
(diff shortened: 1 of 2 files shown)


src/layer/vk_layer_settings.cpp
```diff
@@ -558,8 +558,6 @@ VkResult vlGetLayerSettingValues(VlLayerSettingSet layerSettingSet, const char *
             return result;
         }
     }
-
-    return VK_ERROR_UNKNOWN;
```
juan-lunarg, line 562: I'm guessing this address a compiler warning?

BUILD.gn
juan-lunarg, line 1: Is there CI for this new build?

PR #63, opened 2023-08-15
Add testing for vk_enum_string_helper.h

### LINKED ISSUE #46: Add testing for vk_enum_string_helper.h

juan-lunarg:
Use magic_enum to test vk_enum_string_helper.h

closes #46


tests/generated/vk_enum_string_helper.cpp
```diff
@@ -5,5 +5,72 @@
...
+
+TEST(vk_enum_string_helper, string_VkResult) {
+    constexpr auto values = magic_enum::enum_values<VkResult>();
+    for (auto val : values) {
+        auto magic_str = magic_enum::enum_name(val);
+
+        auto str = string_VkResult(val);
+
+        EXPECT_STREQ(magic_str.data(), str);
+    }
```
juan-lunarg, line 20: https://github.com/Neargye/magic_enum#examples

magic_enum makes testing these functions quite easy.

Note I didn't add testing for each enum possible. I just picked a few significant ones. Since we are mainly testing the code generation above anything else.

spencer-lunarg (approved):
initial reaction: we are grabbing another 3rd party library just to test one of the easier code gen code

after thinking more: 
- its a header only dep (less worry about compiling)
- its only for testing, so people need to opt-in to this anyway
- we should have "some" tests

scripts/generators/enum_string_helper_generator.py
```diff
@@ -31,20 +31,6 @@ def generate(self):
 #include <vulkan/vulkan.h>
 ''')
 
-        # TODO - this should be moved into different generated util file
```
spencer-lunarg, line 34: ... can take action now to move this in VVL so it doesn't break when we switch over to VUL

PR #72, opened 2023-08-23
Remove clang-tidy comments from vk_enum_string_helper.h

juan-lunarg:
Not needed. They only made sense in the context of VVL.


juan-lunarg:
FYI @charles-lunarg remove these comments from the dispatch code as well. Didn't want to do it in this PR to avoid merge conflict.

PR #89, opened 2023-09-01
src: Add vk_format_utils to UtilityHeaders

charles-lunarg:
The library originates from Vulkan-ValidationLayers, but is being moved into this repo to make it easier for others to use it.

The library has also been modified to be header only and C compatible, which allows more developers to be able to use it. This does require some changes, but only affects the FormatElementSize and FormatTexelSize functions which used default parameters. Two new functions, FormatElementSizeWithAspect and FormatTexelSizeWithAspect have been added to handle the non-default image aspect case (the default was COLOR_BIT).

### DIFF
--- a/include/vulkan/utility/vk_format_utils.h
+++ b/include/vulkan/utility/vk_format_utils.h
@@ -1,0 +1,2264 @@
+// *** THIS FILE IS GENERATED - DO NOT EDIT ***
+// See format_utils_generator.py for modifications
+// Copyright 2023 The Khronos Group Inc.
+// Copyright 2023 Valve Corporation
+// Copyright 2023 LunarG, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <vulkan/vulkan.h>
+
+#include <stdbool.h>
+
+#define VKU_FORMAT_INVALID_INDEX 0xFFFFFFFF
+#define VKU_FORMAT_MAX_PLANES 3
+#define VKU_FORMAT_MAX_COMPONENTS 4
+
+enum VKU_FORMAT_NUMERICAL_TYPE {
+    VKU_FORMAT_NUMERICAL_TYPE_NONE = 0,
+    VKU_FORMAT_NUMERICAL_TYPE_SFLOAT,
+    VKU_FORMAT_NUMERICAL_TYPE_SINT,
+    VKU_FORMAT_NUMERICAL_TYPE_SNORM,
+    VKU_FORMAT_NUMERICAL_TYPE_SRGB,
+    VKU_FORMAT_NUMERICAL_TYPE_SSCALED,
+    VKU_FORMAT_NUMERICAL_TYPE_UFLOAT,
+    VKU_FORMAT_NUMERICAL_TYPE_UINT,
+    VKU_FORMAT_NUMERICAL_TYPE_UNORM,
+    VKU_FORMAT_NUMERICAL_TYPE_USCALED,
+};
+
+enum VKU_FORMAT_COMPATIBILITY_CLASS {
+    VKU_FORMAT_COMPATIBILITY_CLASS_NONE = 0,
+    VKU_FORMAT_COMPATIBILITY_CLASS_10BIT_2PLANE_420,
+    VKU_FORMAT_COMPATIBILITY_CLASS_10BIT_2PLANE_422,
+    VKU_FORMAT_COMPATIBILITY_CLASS_10BIT_2PLANE_444,
+    VKU_FORMAT_COMPATIBILITY_CLASS_10BIT_3PLANE_420,
+    VKU_FORMAT_COMPATIBILITY_CLASS_10BIT_3PLANE_422,
+    VKU_FORMAT_COMPATIBILITY_CLASS_10BIT_3PLANE_444,
+    VKU_FORMAT_COMPATIBILITY_CLASS_128BIT,
+    VKU_FORMAT_COMPATIBILITY_CLASS_12BIT_2PLANE_420,
+    VKU_FORMAT_COMPATIBILITY_CLASS_12BIT_2PLANE_422,
+    VKU_FORMAT_COMPATIBILITY_CLASS_12BIT_2PLANE_444,
+    VKU_FORMAT_COMPATIBILITY_CLASS_12BIT_3PLANE_420,
+    VKU_FORMAT_COMPATIBILITY_CLASS_12BIT_3PLANE_422,
+    VKU_FORMAT_COMPATIBILITY_CLASS_12BIT_3PLANE_444,
+    VKU_FORMAT_COMPATIBILITY_CLASS_16BIT,
+    VKU_FORMAT_COMPATIBILITY_CLASS_16BIT_2PLANE_420,
+    VKU_FORMAT_COMPATIBILITY_CLASS_16BIT_2PLANE_422,
+    VKU_FORMAT_COMPATIBILITY_CLASS_16BIT_2PLANE_444,
+    VKU_FORMAT_COMPATIBILITY_CLASS_16BIT_3PLANE_420,
+    VKU_FORMAT_COMPATIBILITY_CLASS_16BIT_3PLANE_422,
+    VKU_FORMAT_COMPATIBILITY_CLASS_16BIT_3PLANE_444,
+    VKU_FORMAT_COMPATIBILITY_CLASS_192BIT,
+    VKU_FORMAT_COMPATIBILITY_CLASS_24BIT,
+    VKU_FORMAT_COMPATIBILITY_CLASS_256BIT,
+    VKU_FORMAT_COMPATIBILITY_CLASS_32BIT,
+    VKU_FORMAT_COMPATIBILITY_CLASS_32BIT_B8G8R8G8,
+    VKU_FORMAT_COMPATIBILITY_CLASS_32BIT_G8B8G8R8,
+    VKU_FORMAT_COMPATIBILITY_CLASS_48BIT,
+    VKU_FORMAT_COMPATIBILITY_CLASS_64BIT,
+    VKU_FORMAT_COMPATIBILITY_CLASS_64BIT_B10G10R10G10,
+    VKU_FORMAT_COMPATIBILITY_CLASS_64BIT_B12G12R12G12,
+    VKU_FORMAT_COMPATIBILITY_CLASS_64BIT_B16G16R16G16,
+    VKU_FORMAT_COMPATIBILITY_CLASS_64BIT_G10B10G10R10,
+    VKU_FORMAT_COMPATIBILITY_CLASS_64BIT_G12B12G12R12,
+    VKU_FORMAT_COMPATIBILITY_CLASS_64BIT_G16B16G16R16,
+    VKU_FORMAT_COMPATIBILITY_CLASS_64BIT_R10G10B10A10,
+    VKU_FORMAT_COMPATIBILITY_CLASS_64BIT_R12G12B12A12,
+    VKU_FORMAT_COMPATIBILITY_CLASS_8BIT,
+    VKU_FORMAT_COMPATIBILITY_CLASS_8BIT_2PLANE_420,
+    VKU_FORMAT_COMPATIBILITY_CLASS_8BIT_2PLANE_422,
+    VKU_FORMAT_COMPATIBILITY_CLASS_8BIT_2PLANE_444,
+    VKU_FORMAT_COMPATIBILITY_CLASS_8BIT_3PLANE_420,
+    VKU_FORMAT_COMPATIBILITY_CLASS_8BIT_3PLANE_422,
+    VKU_FORMAT_COMPATIBILITY_CLASS_8BIT_3PLANE_444,
+    VKU_FORMAT_COMPATIBILITY_CLASS_8BIT_ALPHA,
+    VKU_FORMAT_COMPATIBILITY_CLASS_96BIT,
+    VKU_FORMAT_COMPATIBILITY_CLASS_ASTC_10X10,
+    VKU_FORMAT_COMPATIBILITY_CLASS_ASTC_10X5,
+    VKU_FORMAT_COMPATIBILITY_CLASS_ASTC_10X6,
+    VKU_FORMAT_COMPATIBILITY_CLASS_ASTC_10X8,
+    VKU_FORMAT_COMPATIBILITY_CLASS_ASTC_12X10,
+    VKU_FORMAT_COMPATIBILITY_CLASS_ASTC_12X12,
+    VKU_FORMAT_COMPATIBILITY_CLASS_ASTC_4X4,
+    VKU_FORMAT_COMPATIBILITY_CLASS_ASTC_5X4,
+    VKU_FORMAT_COMPATIBILITY_CLASS_ASTC_5X5,
+    VKU_FORMAT_COMPATIBILITY_CLASS_ASTC_6X5,
+    VKU_FORMAT_COMPATIBILITY_CLASS_ASTC_6X6,
+    VKU_FORMAT_COMPATIBILITY_CLASS_ASTC_8X5,
+    VKU_FORMAT_COMPATIBILITY_CLASS_ASTC_8X6,
+    VKU_FORMAT_COMPATIBILITY_CLASS_ASTC_8X8,
+    VKU_FORMAT_COMPATIBILITY_CLASS_BC1_RGB,
+    VKU_FORMAT_COMPATIBILITY_CLASS_BC1_RGBA,
+    VKU_FORMAT_COMPATIBILITY_CLASS_BC2,
+    VKU_FORMAT_COMPATIBILITY_CLASS_BC3,
+    VKU_FORMAT_COMPATIBILITY_CLASS_BC4,
+    VKU_FORMAT_COMPATIBILITY_CLASS_BC5,
+    VKU_FORMAT_COMPATIBILITY_CLASS_BC6H,
+    VKU_FORMAT_COMPATIBILITY_CLASS_BC7,
+    VKU_FORMAT_COMPATIBILITY_CLASS_D16,
+    VKU_FORMAT_COMPATIBILITY_CLASS_D16S8,
+    VKU_FORMAT_COMPATIBILITY_CLASS_D24,
+    VKU_FORMAT_COMPATIBILITY_CLASS_D24S8,
+    VKU_FORMAT_COMPATIBILITY_CLASS_D32,
+    VKU_FORMAT_COMPATIBILITY_CLASS_D32S8,
+    VKU_FORMAT_COMPATIBILITY_CLASS_EAC_R,
+    VKU_FORMAT_COMPATIBILITY_CLASS_EAC_RG,
+    VKU_FORMAT_COMPATIBILITY_CLASS_ETC2_EAC_RGBA,
+    VKU_FORMAT_COMPATIBILITY_CLASS_ETC2_RGB,
+    VKU_FORMAT_COMPATIBILITY_CLASS_ETC2_RGBA,
+    VKU_FORMAT_COMPATIBILITY_CLASS_PVRTC1_2BPP,
+    VKU_FORMAT_COMPATIBILITY_CLASS_PVRTC1_4BPP,
+    VKU_FORMAT_COMPATIBILITY_CLASS_PVRTC2_2BPP,
+    VKU_FORMAT_COMPATIBILITY_CLASS_PVRTC2_4BPP,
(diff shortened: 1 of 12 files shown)


tests/add_subdirectory/vk_format_utils.c
```diff
@@ -0,0 +1,8 @@
+// Copyright 2023 The Khronos Group Inc.
+// Copyright 2023 Valve Corporation
+// Copyright 2023 LunarG, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+#include <vulkan/utility/vk_format_utils.h>
+
+void foo() {}
```
spencer-lunarg, line 8: what is this for?

include/vulkan/utility/vk_format_utils.h
```diff
@@ -0,0 +1,2297 @@
...
+    FORMAT_UTILS_NUMERICAL_TYPE_SNORM,
+    FORMAT_UTILS_NUMERICAL_TYPE_SRGB,
+    FORMAT_UTILS_NUMERICAL_TYPE_SSCALED,
+    FORMAT_UTILS_NUMERICAL_TYPE_UFLOAT,
+    FORMAT_UTILS_NUMERICAL_TYPE_UINT,
+    FORMAT_UTILS_NUMERICAL_TYPE_UNORM,
+    FORMAT_UTILS_NUMERICAL_TYPE_USCALED
+};
+
+enum FORMAT_UTILS_COMPATIBILITY_CLASS {
```
spencer-lunarg, line 35: 1. would it be better to forward declare so people looking at the file can quickly see the functions at the top?
2. I guess curious if this should be `VK_FORMAT_UTILS_*`... should we use `VK_FORMAT_UTILS` name space to match the filename

spencer-lunarg (changes requested):
I guess before we merge here, a good exercise would be go in the Validation Layers

1. Remove the current format utils (so it doesn't generate)
2. Replace the header with this file
3. Make sure CI will pass and this didn't forget anything subtle a test there would have caught

(the goal not being to merge that PR, but to prove this new interface works)

tests/add_subdirectory/vk_format_utils.c
charles-lunarg: This tests whether the code is actually valid C code - since I just need to include it for the compiler to check, I should probably rename the function and add some usage just for testings sake.

include/vulkan/utility/vk_format_utils.h
charles-lunarg: 1. Forward declaring the enums is a good idea, I made the functions all forward declared for the same logic after all. 

2. VK_ looks like a vulkan api enum which I do not want to be mistaken as.

charles-lunarg:
> I guess before we merge here, a good exercise would be go in the Validation Layers

I started to do that on Friday and did not have much trouble with the conversion but had a lot of trouble running tests (w and w/o mock icd). It felt like I setup the repo wrong but I was on a different machine than I normally develop on, so I’ll try again come the morn on that machine.

tests/add_subdirectory/vk_format_utils.c
artem-lunarg: Is it possible to make this file "ephemeral", by generating only during the build phase? If that makes sense in this context. I guess this method is used by some configuration tools that generate small files to check for support of particular platform/language feature.

include/vulkan/utility/vk_format_utils.h
```diff
@@ -0,0 +1,2297 @@
...
+    FORMAT_UTILS_COMPATIBILITY_CLASS_PVRTC2_2BPP,
+    FORMAT_UTILS_COMPATIBILITY_CLASS_PVRTC2_4BPP,
+    FORMAT_UTILS_COMPATIBILITY_CLASS_S8
+};
+// Return the plane index of a given VkImageAspectFlagBits.
+//     VK_IMAGE_ASPECT_PLANE_0_BIT -> 0
+//     VK_IMAGE_ASPECT_PLANE_1_BIT -> 1
+//     VK_IMAGE_ASPECT_PLANE_2_BIT -> 2
+//     <any other value> -> FORMAT_UTILS_MAX_PLANES
+inline uint32_t GetPlaneIndex(VkImageAspectFlagBits aspect);
```
artem-lunarg, line 126: One option is to return uint_max here (and to rename constant to something with INVALID or similar). Invalid program has more chances not to crash with value 3 (which is close to correct values) than with a large value. Another consideration point is to which degree should we care about backward compatibility. Using large invalid constant will help the library to be less affected by some future PLANE_3_BIT (not sure how probably is it).

include/vulkan/utility/vk_format_utils.h
```diff
@@ -0,0 +1,2297 @@
...
+    FORMAT_UTILS_NUMERICAL_TYPE_NONE = 0,
+    FORMAT_UTILS_NUMERICAL_TYPE_SFLOAT,
+    FORMAT_UTILS_NUMERICAL_TYPE_SINT,
+    FORMAT_UTILS_NUMERICAL_TYPE_SNORM,
+    FORMAT_UTILS_NUMERICAL_TYPE_SRGB,
+    FORMAT_UTILS_NUMERICAL_TYPE_SSCALED,
+    FORMAT_UTILS_NUMERICAL_TYPE_UFLOAT,
+    FORMAT_UTILS_NUMERICAL_TYPE_UINT,
+    FORMAT_UTILS_NUMERICAL_TYPE_UNORM,
+    FORMAT_UTILS_NUMERICAL_TYPE_USCALED
```
artem-lunarg, line 32: nit: suggestion to end last enum line with a comma, to remove corresponding line from the future diffs if new members are added.

include/vulkan/utility/vk_format_utils.h
```diff
@@ -0,0 +1,2297 @@
...
+    FORMAT_UTILS_COMPATIBILITY_CLASS_S8
+};
+// Return the plane index of a given VkImageAspectFlagBits.
+//     VK_IMAGE_ASPECT_PLANE_0_BIT -> 0
+//     VK_IMAGE_ASPECT_PLANE_1_BIT -> 1
+//     VK_IMAGE_ASPECT_PLANE_2_BIT -> 2
+//     <any other value> -> FORMAT_UTILS_MAX_PLANES
+inline uint32_t GetPlaneIndex(VkImageAspectFlagBits aspect);
+
+// Returns whether a format is of the numberical type SFLOAT
```
artem-lunarg, line 128: typo: numberical

include/vulkan/utility/vk_format_utils.h
```diff
@@ -0,0 +1,2297 @@
...
+
+inline bool FormatsEqualComponentBits(VkFormat format_a, VkFormat format_b) {
+    const struct FORMAT_UTILS_FORMAT_INFO format_info_a = GetFormatInfo(format_a);
+    const struct FORMAT_UTILS_FORMAT_INFO format_info_b = GetFormatInfo(format_b);
+    if (format_info_a.compatibility == FORMAT_UTILS_COMPATIBILITY_CLASS_NONE || format_info_b.compatibility == FORMAT_UTILS_COMPATIBILITY_CLASS_NONE) {
+        return false;
+    } else if (format_info_a.component_count != format_info_b.component_count) {
+        return false;
+    }
+    // Need to loop match each component type is found in both formats
```
artem-lunarg, line 2259: Why don't we match just corresponding components? Matching based on possible permutations is not intuitive and is not something I'd would expect from api like this. Matching based on direct permutations looks like more practical use case.

include/vulkan/utility/vk_format_utils.h
```diff
@@ -0,0 +1,2297 @@
...
+// Returns whether a format is of the numberical type UNORM
+// Format must only contain one numberical type, so formats like D16_UNORM_S8_UINT always return false
+inline bool FormatIsUNORM(VkFormat format);
+
+// Returns whether a format is of the numberical type USCALED
+// Format must only contain one numberical type, so formats like D16_UNORM_S8_UINT always return false
+inline bool FormatIsUSCALED(VkFormat format);
+
+// Returns whether a format type is any type from "Interpretation of Numeric Format" table (OpTypeInt)
+inline bool FormatIsSampledInt(VkFormat format);
```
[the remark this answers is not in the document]
charles-lunarg: I wasn't super clear on that myself - but I'll look at the format table in question and double check that.

include/vulkan/utility/vk_format_utils.h
```diff
@@ -0,0 +1,2297 @@
...
+
+inline bool FormatHasRed(VkFormat format) { return FormatHasComponentType(format, FORMAT_UTILS_COMPONENT_TYPE_R); }
+
+inline bool FormatHasGreen(VkFormat format) { return FormatHasComponentType(format, FORMAT_UTILS_COMPONENT_TYPE_G); }
+
+inline bool FormatHasBlue(VkFormat format) { return FormatHasComponentType(format, FORMAT_UTILS_COMPONENT_TYPE_B); }
+
+inline bool FormatHasAlpha(VkFormat format) { return FormatHasComponentType(format, FORMAT_UTILS_COMPONENT_TYPE_A); }
+
+inline bool FormatsEqualComponentBits(VkFormat format_a, VkFormat format_b) {
```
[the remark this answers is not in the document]
charles-lunarg: FormatsIsEqualComponentBits sets off all sorts of 'grammar' warning bells in my head, as it should be FormatsAreEqualComponentBits... (singular vs plural).

Naming ~~are~~ is hard!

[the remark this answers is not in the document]
artem-lunarg: 💯

include/vulkan/utility/vk_format_utils.h
charles-lunarg: So I didn't write the code, I just fixed it up for this PR. 

My understanding is that this is directly checking permutations of formats. R8G8B8A8 == B8G8R8A8 would be an example. Checking that R8G8B8A8_UNORM == R8G8B8A8_SRGB would require a different function, and I'm not sure vk_format_utils.h has anything for that currently.

include/vulkan/utility/vk_format_utils.h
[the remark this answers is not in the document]
spencer-lunarg: so for the formats, awhile ago I tried to unify the wording

https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#formats-numericformat

[internal MR](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/5627)

in short:

- **Numeric Format** = [`SFLOAT`, `SINT`, `SNORM`, `SRGB`, `SSCALED`, `UFLOAT`, `UINT`, `UNORM`, `USCALED`]
- **Numeric Type** = [float, signed int, unsigned int]
- **SPIR-V Type** = [ [`OpTypeFloat`, `OpTypeInt`], width, sign ]

.... basically we should re-write this comment to whatever @charles-lunarg  thinks "makes sense" as I am now to tainted with spec lawyer knowledge

include/vulkan/utility/vk_format_utils.h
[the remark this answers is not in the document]
spencer-lunarg: I don't care what it is called, just that everything else is also consistently the same

> Naming is hard

:100:

include/vulkan/utility/vk_format_utils.h
spencer-lunarg: the only use of this is in `VUID-VkImageViewCreateInfo-imageViewFormatReinterpretation-04466`

I needed a way to check the `R, G, B` != `R, G, A,` and only way I could think of was a double for-loop

charles-lunarg: Is this an example of a function which is useful for VVL and thats about it? Not sure how to use this function in not-validation code.

spencer-lunarg: this is an example of where maybe the answer is 

"Have a function that returns `COMPONENT_INFO` and have the end users do the checking.... Like having `FormatsSameComponentBits` in here makes it easy to justify having so many other random util checks in as well that most people don't need.... I am not sure what level of abstraction for a Format Util header is useful, but this is on the edge of "very specific" use case IMO

charles-lunarg: Since this is a header library, all those tools definitely exist. Before, it was all hidden in the .cpp file so users couldn't just get the info struct and compare the components themselves. But now? I would like to pursue removing this function and make VVL do what it needs (I'll of course implement that logic when I get around to making VVL use the format utils header).

Not sure what other functions in the header have a similar concern, as most everything takes a single format instead of 2 and most everything can't be easily checked themselves (lots of logic that benefits from codegen while this function does not).

tests/add_subdirectory/vk_format_utils.c
juan-lunarg: It not only tests the C code is valid.

It helps test add_subdirectory / find_package utilization.

juan-lunarg: IE it's integration testing to make sure we are writing good CMake for users to absorb our libaries

tests/format_utils/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+# Copyright 2023 The Khronos Group Inc.
+# Copyright 2023 Valve Corporation
+# Copyright 2023 LunarG, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+find_package(GTest REQUIRED CONFIG)
+find_package(magic_enum REQUIRED CONFIG)
+
+include(GoogleTest)
```
juan-lunarg, line 10: Note I don't like how this code is repeated in our CMake testing code. I'll make a fix for this in another PR though. It's currently consistent as is.

juan-lunarg:
Did some minimal integration testing with VulkanTools. Looks good on that front 👍🏾 
https://github.com/LunarG/VulkanTools/pull/1883

include/vulkan/utility/vk_format_utils.h
```diff
@@ -0,0 +1,2297 @@
...
+//     VK_IMAGE_ASPECT_PLANE_2_BIT -> 2
+//     <any other value> -> FORMAT_UTILS_MAX_PLANES
+inline uint32_t GetPlaneIndex(VkImageAspectFlagBits aspect);
+
+// Returns whether a format is of the numberical type SFLOAT
+// Format must only contain one numberical type, so formats like D16_UNORM_S8_UINT always return false
+inline bool FormatIsSFLOAT(VkFormat format);
+
+// Returns whether a format is of the numberical type SINT
+// Format must only contain one numberical type, so formats like D16_UNORM_S8_UINT always return false
```
juan-lunarg, line 133:
```suggestion
// Format must only contain one numerical type, so formats like D16_UNORM_S8_UINT always return false
```

include/vulkan/utility/vk_format_utils.h
```diff
@@ -0,0 +1,2297 @@
...
+
+// Return the size in bytes of one texel of a foramt
+// For compressed or multi-plane, this may be a fractional number
+inline uint32_t FormatElementSizeWithAspect(VkFormat format, VkImageAspectFlagBits aspectMask);
+
+// Return the size in bytes of one texel of a foramt
+// Format must not be a depth, stencil, or multiplane format
+inline double FormatTexelSize(VkFormat format);
+
+// Return the size in bytes of one texel of a foramt
```
juan-lunarg, line 278:
```suggestion
// Return the size in bytes of one texel of a format
```

tests/format_utils/test_formats.cpp
```diff
@@ -0,0 +1,681 @@
...
+    return comp_count;
+}
+
+TEST(FormatUtils, FormatIsSFLOAT) {
+    for (auto [format, format_str] : magic_enum::enum_entries<VkFormat>()) {
+        // special case depth + stencil formats
+        if (format == VK_FORMAT_D32_SFLOAT_S8_UINT) {
+            EXPECT_FALSE(FormatIsSFLOAT(format));
+            continue;
+        }
```
juan-lunarg, line 60: Is the rational because the stencil isn't SFLOAT? IE this function only returns true if each component of the format is SFLOAT?

charles-lunarg: This was how the function worked before, I do not know how rational such a decision is. I guess it makes sense, but I think asking who ever added the function originally would be the best person.

juan-lunarg: It's documented so it's fine. Perhaps having a FormatHasSFLOAT would be appropriate to have. But that a later feature request for another PR.

spencer-lunarg: I added this, there are some strange edge cases where one of these dual Depth/Stencil formats was returning true and we didn't want that behavior... I guess the `FormatHas` is just this but would return true for `D32S8` but not sure a use case... like if you are trying to match your Shader Interface, these dual formats are not allowed anyway

juan-lunarg: >  but not sure a use case...

Right. I'm not sure there is a use case either. Which is why it's fine as is. If we need more functionality we can add it later.

include/vulkan/utility/vk_format_utils.h
juan-lunarg: >  1. would it be better to forward declare so people looking at the file can quickly see the functions at the top?

I personally disagree on forward declaring. Most IDEs offer a way to collapse all functions/scopes. So IMO the forward declarations just adds more clutter/complexity.

juan-lunarg: > VK_ looks like a vulkan api enum which I do not want to be mistaken as.

FFTI. I think using `VUL_` would be more appropriate.

scripts/generators/format_utils_generator.py
```diff
@@ -0,0 +1,695 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+from generators.vulkan_object import (Format)
+from generators.base_generator import BaseGenerator
+
+# Make C++ name friendly class name
+def getClassName(className: str) -> str:
+    return className.replace('-', '').replace(' ', '_').upper()
```
juan-lunarg, line 15: Is this needed? We are generating a C header.

charles-lunarg: Oh absolutely - its 'C++ & C' friendly.

Spec uses `16-bit 2-plane 420` whereas we need `16BIT_2PLANE_420`.

include/vulkan/utility/vk_format_utils.h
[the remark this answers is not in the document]
charles-lunarg: Let me know if the comments make more sense now.

include/vulkan/utility/vk_format_utils.h
charles-lunarg: Added FORMAT_UTILS_INVALID_INDEX = 0xFFFFFFFF to make it clearer. The MAX_COMPONENTS is required since that creates the size of the components array.

include/vulkan/utility/vk_format_utils.h
charles-lunarg: Forward declaring can be annoying since `enum` doesn't declare a size, but sizes are needed for forward declarations. I am not going to try to do that here, since I don't want to delay this PR any further while I wrestle getting enum forward decls working in C++ & C.

include/vulkan/utility/vk_format_utils.h
```diff
@@ -0,0 +1,2268 @@
...
+
+inline bool FormatHasBlue(VkFormat format) { return FormatHasComponentType(format, FORMAT_UTILS_COMPONENT_TYPE_B); }
+
+inline bool FormatHasAlpha(VkFormat format) { return FormatHasComponentType(format, FORMAT_UTILS_COMPONENT_TYPE_A); }
+
+inline uint32_t GetPlaneIndex(VkImageAspectFlagBits aspect) {
+    switch (aspect) {
+        case VK_IMAGE_ASPECT_PLANE_0_BIT:
+            return 0;
+            break;
```
ziga-lunarg, line 2253: `break` after `return`?

charles-lunarg: Originally it didn't return from inside the switch statement, but now that it does, the break is redundant. Thanks for catching it!

juan-lunarg (changes requested):
This will require updating `BUILD.gn`

include/vulkan/utility/vk_format_utils.h
juan-lunarg: Agreed that's why I said FFTI.

juan-lunarg (approved):

charles-lunarg:
Apologies for the force push spam - the C compatibility tests & add_subdirectory/find_package tests are easy to forget about since they aren't compiled normally.

The build.gn file should be correct now as well @juan-lunarg

juan-lunarg (approved):
LGTM

PR #95, opened 2023-09-05
ci: Add clang-tidy to CI

### LINKED ISSUE #76: Run clang-tidy on header files that we ship
Due to CMake's built in clang-tidy functionality only working on Ninja / Makefile generators using the built in functionality is problematic.

However, it's still straightforward to make sure that at the least the files we are installing follow proper conventions and coding guidelines.

IE the header files we ship will be consistent in quality. With the exception of vk_enum_string_helper.h due to legacy reasons.

juan-lunarg:
closes #76

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -42,8 +42,6 @@ cd build/
 ctest -C Debug --parallel 8 --output-on-failure
 ```
 
-## CMake
-
 ### Warnings as errors off by default!
 
 By default `VUL_WERROR` is `OFF`
@@ -55,3 +53,9 @@ System package managers, and language package managers have to build on multiple
 By defaulting to `ON` we cause issues for package managers since there is no standard way to disable warnings until CMake 3.24
 
 Add `-D VUL_WERROR=ON` to your workflow. Or use the `dev` preset shown below which will also enabling warnings as errors.
+
+### clang-tidy
+
+By default clang-tidy checking is disabled by default.
+
+To enable it set `VUL_CLANG_TIDY` to `ON`. clang-tidy is required to pass CI.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,15 @@ if (VUL_WERROR)
     endif()
 endif()
 
+option(VVL_CLANG_TIDY "Use clang-tidy")
+if (VUL_CLANG_TIDY)
+    find_program(CLANG_TIDY NAMES clang-tidy)
+    if(NOT CLANG_TIDY)
+        message(FATAL_ERROR "clang-tidy not found!")
+    endif()
+    set(CMAKE_CXX_CLANG_TIDY "${CLANG_TIDY}")
+endif()
+
 if(${CMAKE_CXX_COMPILER_ID} MATCHES "(GNU|Clang)")
     add_compile_options(
         -Wall
--- a/src/layer/layer_settings_manager.cpp
+++ b/src/layer/layer_settings_manager.cpp
@@ -79,7 +79,7 @@ static inline bool IsHighIntegrity() {
         DWORD buffer_size;
         if (GetTokenInformation(process_token, TokenIntegrityLevel, mandatory_label_buffer, sizeof(mandatory_label_buffer),
                                 &buffer_size) != 0) {
-            const TOKEN_MANDATORY_LABEL *mandatory_label = (const TOKEN_MANDATORY_LABEL *)mandatory_label_buffer;
+            const TOKEN_MANDATORY_LABEL *mandatory_label = reinterpret_cast<const TOKEN_MANDATORY_LABEL *>(mandatory_label_buffer);
             const DWORD sub_authority_count = *GetSidSubAuthorityCount(mandatory_label->Label.Sid);
             const DWORD integrity_level = *GetSidSubAuthority(mandatory_label->Label.Sid, sub_authority_count - 1);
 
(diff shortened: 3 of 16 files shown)


src/layer/layer_settings_util.cpp
```diff
@@ -230,34 +230,34 @@ std::vector<VlFrameset> ToFrameSets(const std::string &s) {
...
-    static const std::regex FRAME_REGEX("^-?[0-9]*([.][0-9]*f?)?$");
+    static const std::regex frame_regex("^-?[0-9]*([.][0-9]*f?)?$");
 
-    return std::regex_search(s, FRAME_REGEX);
+    return std::regex_search(s, frame_regex);
 }
 
 std::string FormatString(const char *message, ...) {
-    std::size_t const STRING_BUFFER(4096);
+    std::size_t const string_buffer(4096);
```
charles-lunarg, line 251: string_buffer is a confusing name since its the *length* of the buffer, not the buffer itself.

charles-lunarg (approved):
Looks like the minimal changes needed to make clang-tidy work *and* fix all the naming violations. 

If this somehow gets merged before format utils, I better rebase and fix any incipient errors

BUILD.md
```diff
@@ -55,3 +53,9 @@ System package managers, and language package managers have to build on multiple
 By defaulting to `ON` we cause issues for package managers since there is no standard way to disable warnings until CMake 3.24
 
 Add `-D VUL_WERROR=ON` to your workflow. Or use the `dev` preset shown below which will also enabling warnings as errors.
+
+### clang-tidy
+
+By default clang-tidy checking is disabled by default.
```
christophe-lunarg, line 59: Why is it disabled by default? It seems much better for me to hit issues on my system than on C.I., consume more C.I. resource and have to iterate, taking more development time. This is especially true that I don't see how the developer would even now what the rules are.

I would argu the same with VUL_WERROR being OFF by default.

I think we need to be more consious when we add new checks how this is going to affect the developers productivity.

src/layer/layer_settings_manager.cpp
```diff
@@ -150,17 +150,17 @@ std::filesystem::path LayerSettings::FindSettingsFile() {
         LSTATUS err = RegOpenKeyEx(hives[hive_index], TEXT("Software\\Khronos\\Vulkan\\Settings"), 0, KEY_READ, &key);
         if (err == ERROR_SUCCESS) {
             TCHAR name[2048];
-            DWORD i = 0, name_size, type, pValues, value_size;
+            DWORD i = 0, name_size, type, p_values, value_size;
```
christophe-lunarg, line 153: I have never seen this convention before.

christophe-lunarg (commented):
I am personnally not in favour of such changes. C.I. making more checks that on local system by default is hunting productivity. And if to increase productivity we need to use the flag, then the flag should be on my default.

Do we have in order already the -WERROR things on repo? Meaning that if C.I. is using it, we should use in by default on our system.

This case is already pretty painful to me, I don't want to deal with another case like this that in this end provide little value expect avoiding pedentic disucssions in code reviews.

Personnally I think that if we add such things, this should be discuss in engineering meetings with the team.

src/layer/layer_settings_manager.cpp
juan-lunarg: I grabbed the clang-tidy file from the vulkan validation layers.

BUILD.md
juan-lunarg: It's off by default to avoid issues with package managers, machines that don't have clang-tidy installed by default, FetchContent/add_subdirectory users, etc.

Having it on by default is more complicated than having it on. That's just the way it is with open source projects.

src/layer/layer_settings_manager.cpp
juan-lunarg: I have no preference tbh. Just a matter of changing the clang-tidy file.

```
- { key: readability-identifier-naming.VariableCase, value: lower_case }
```

juan-lunarg:
> Why is it disabled by default?

Another reason to also have it disabled by default is because if your IDE registers .clang-tidy files. (Visual Studio Code, Visual Studio, and many other IDEs do). Is that it will pointlessly eat up build time. Since you can just fix the clang-tidy issues in your IDE.

I've looked at various other repos that have implemented clang-tidy and this seems to be the convention from what I can gather.

> Personnally I think that if we add such things, this should be discuss in engineering meetings with the team.

Understood. This has been brought up multiple times in VVL meetings. It's a lot harder to implement on an existing codebase however. Since it would require a lot of fixes. The thought process was to get it working on a newer repo to iron the kinks in the workflow.

> little value expect avoiding pedentic disucssions in code reviews.

As you said. The main intent behind clang-tidy is to enforce a style guide essentially. It's very flexible/powerful.

However, it does more than just style checking. It provides performance static analysis checks, security checks, etc.

juan-lunarg:
To be clear. I do understand the annoyance of adding more checks. But overall I thinks it's better for the health of the multiple code bases we have to manage.

juan-lunarg:
I've got a better idea that would go more smoothly. Closing this PR.

CMakeLists.txt
```diff
@@ -31,6 +31,15 @@ if (VUL_WERROR)
     endif()
 endif()
 
+option(VVL_CLANG_TIDY "Use clang-tidy")
+if (VUL_CLANG_TIDY)
+    find_program(CLANG_TIDY NAMES clang-tidy)
+    if(NOT CLANG_TIDY)
+        message(FATAL_ERROR "clang-tidy not found!")
+    endif()
+    set(CMAKE_CXX_CLANG_TIDY "${CLANG_TIDY}")
```
juan-lunarg, line 40: CMAKE_\<LANG\>_CLANG_TIDY only works on the makefile / ninja generator.

Given how popular the visual studio generator is within our team this is likely only going to cause confusion.

PR #98, opened 2023-09-06
cmake: Add a function to setup all build targets with the same strictness

christophe-lunarg:
Question: Here I am using `add_compile_options` does it affect other target that would add this repository to it's project? 
Using `find_package` and `add_subdirctory`?

I am using `add_compile_options` because this is was used for `/WX`.


juan-lunarg:
> Question: Here I am using add_compile_options does it affect other target that would add this repository to it's project?
Using find_package and add_subdirctory?

`add_compile_options` won't affect projects found via find_package. That's one of the reasons I actually prefer find_package.

`add_compile_options` WILL affect the project absorbed using `add_subdirectory`. This is often annoying when dealing with third party libraries.

CMakeLists.txt
```diff
@@ -31,12 +31,7 @@ if (VUL_WERROR)
     endif()
 endif()
 
-if(${CMAKE_CXX_COMPILER_ID} MATCHES "(GNU|Clang)")
-    add_compile_options(
-        -Wall
-        -Wextra
-    )
-endif()
+add_compile_options("$<IF:$<CXX_COMPILER_ID:MSVC>,/W4,-Wall -Wextra>")
```
juan-lunarg, line 34: I'd prefer we make it consistent with VVL. Like the following.

```suggestion
if(${CMAKE_CXX_COMPILER_ID} MATCHES "(GNU|Clang)")
    add_compile_options(
        -Wall
        -Wextra
        -Wpointer-arith
    )
    if(${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
        add_compile_options(
            -Wconversion
            -Wimplicit-fallthrough
            -Wstring-conversion
        )
    endif()
elseif(MSVC)
    add_compile_options(
        /W4
        /we5038 # Enable warning about MIL ordering in constructors
    )

    # Enforce stricter ISO C++
    add_compile_options(/permissive-)

    # Allow usage of unsafe CRT functions and minimize what Windows.h leaks
    add_compile_definitions(_CRT_SECURE_NO_WARNINGS NOMINMAX WIN32_LEAN_AND_MEAN)

    add_compile_options($<$<BOOL:${MSVC_IDE}>:/MP>) # Speed up Visual Studio builds
```

christophe-lunarg: So maybe we should use `target_compile_options` and `target_compile_definitions` across repositories to avoid issues with `add_subdirectory` cases?

juan-lunarg (approved):

scripts/CMakeLists.txt
```diff
@@ -101,6 +101,47 @@ if (UPDATE_DEPS)
         include("${UPDATE_DEPS_DIR}/helper.cmake")
     endif()
 endif()
+
+# Configure compiler build options common to all targets
+function(lunarg_target_compiler_configurations TARGET_NAME WERROR_OPTION)
```
juan-lunarg, line 106: What would be ideal is if we could actually share this code elegantly with the other repos that consume the utility libraries.

I think this can be done. I'll wait until this gets merged though.

christophe-lunarg: Yes that's what I was thinking!

PR #104, opened 2023-09-11
cmake: Fix accidentally exporting warnings

### LINKED ISSUE #103: UtilityLibraries accidentally exported stricter warnings
`VulkanUtilityLibrariesConfig.cmake` contains the following now.

```
# Create imported target Vulkan::LayerSettings
add_library(Vulkan::LayerSettings STATIC IMPORTED)

set_target_properties(Vulkan::LayerSettings PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "_CRT_SECURE_NO_WARNINGS;NOMINMAX;WIN32_LEAN_AND_MEAN"
  INTERFACE_COMPILE_OPTIONS "/W4;/we5038;/permissive-;\$<\$<BOOL:0>:/MP>"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "Vulkan::Headers"
)

# Create imported target Vulkan::UtilityHeaders
add_library(Vulkan::UtilityHeaders INTERFACE IMPORTED)

set_target_properties(Vulkan::UtilityHeaders PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "_CRT_SECURE_NO_WARNINGS;NOMINMAX;WIN32_LEAN_AND_MEAN"
  INTERFACE_COMPILE_OPTIONS "/W4;/we5038;/permissive-;\$<\$<BOOL:0>:/MP>"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "Vulkan::Headers"
)
```

This was added in https://github.com/KhronosGroup/Vulkan-Utility-Libraries/commit/a3235b3e7fcef4a266173c611086aa1a216856b1 by @christophe-lunarg.

FYI @charles-lunarg

juan-lunarg:
closes #103

### DIFF
--- a/scripts/CMakeLists.txt
+++ b/scripts/CMakeLists.txt
@@ -104,13 +104,15 @@ endif()
 
 # Configure compiler build options common to all targets
 function(lunarg_target_compiler_configurations TARGET_NAME WERROR_OPTION)
-    if (${WERROR_OPTION})
-        if (MSVC)
-            target_compile_options(${TARGET_NAME} INTERFACE /WX)
-            target_link_options(${TARGET_NAME} INTERFACE /WX)
-        else()
-            target_compile_options(${TARGET_NAME} INTERFACE -Werror)
-        endif()
+    if (NOT ${WERROR_OPTION})
+        return()
+    endif()
+
+    if (MSVC)
+        target_compile_options(${TARGET_NAME} INTERFACE /WX)
+        target_link_options(${TARGET_NAME} INTERFACE /WX)
+    else()
+        target_compile_options(${TARGET_NAME} INTERFACE -Werror)
     endif()
 
     if(${CMAKE_CXX_COMPILER_ID} MATCHES "(GNU|Clang)")


charles-lunarg (approved):

christophe-lunarg:
@juan-lunarg, I don't understand the fix. How this was exporting warnings? This should only affect the targets that call `lunarg_target_compiler_configurations` no?

And the fix implies that if we don't enable WERROR when we don't get the warnings at all... That's really wasn't the intention of the function.

juan-lunarg:
> @juan-lunarg, I don't understand the fix. How this was exporting warnings? This should only affect the targets that call `lunarg_target_compiler_configurations` no?

If you look at the `VulkanUtilityLibrariesConfig.cmake` it was exporting compiler options. This is due to the usage of `INTERFACE` in `lunarg_target_compiler_configurations`.

```
# Create imported target Vulkan::LayerSettings
add_library(Vulkan::LayerSettings STATIC IMPORTED)

set_target_properties(Vulkan::LayerSettings PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "_CRT_SECURE_NO_WARNINGS;NOMINMAX;WIN32_LEAN_AND_MEAN"
  INTERFACE_COMPILE_OPTIONS "/W4;/we5038;/permissive-;\$<\$<BOOL:0>:/MP>"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "Vulkan::Headers"
)

# Create imported target Vulkan::UtilityHeaders
add_library(Vulkan::UtilityHeaders INTERFACE IMPORTED)

set_target_properties(Vulkan::UtilityHeaders PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "_CRT_SECURE_NO_WARNINGS;NOMINMAX;WIN32_LEAN_AND_MEAN"
  INTERFACE_COMPILE_OPTIONS "/W4;/we5038;/permissive-;\$<\$<BOOL:0>:/MP>"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "Vulkan::Headers"
)
```

christophe-lunarg:
_If you look at the VulkanUtilityLibrariesConfig.cmake it was exporting compiler options. This is due to the usage of INTERFACE in lunarg_target_compiler_configurations._

@juan-lunarg I guess my issue here is to what the compiler options are exported. But it's probably because I am not clear on expactly what PUBLIC, INTERFACE and PRIVATE really means.

And then if the issue what the INTERFACE value, why the commit do a return instead of using the correct value?

juan-lunarg:
> But it's probably because I am not clear on expactly what PUBLIC, INTERFACE and PRIVATE really means.

Here are some articles going into detail explaining that concept:
- https://kubasejdak.com/modern-cmake-is-like-inheritance
- https://leimao.github.io/blog/CMake-Public-Private-Interface/

christophe-lunarg:
@juan-lunarg so if we use INTERFACE, the effect is said to be:
_CMake 3.x introduced a very important “side effect” of linking with targets: linked target is passing all its PUBLIC and INTERFACE properties to the library that it links with. `_

Then in that case, any project linking with the VulkanUtilitiesHeader or VulkanLayerSettings would get the build settings configuration. That's sounds like what we want no?

Actually with static libraries that mostly required too for some settings...

christophe-lunarg:
"Modern CMake is like inheritance" That title was quite an epiphany!

Do I like that? No. But I get it.

juan-lunarg:
> Then in that case, any project linking with the VulkanUtilitiesHeader or VulkanLayerSettings would get the build settings configuration. That's sounds like what we want no?

It's a [side effect](https://en.wikipedia.org/wiki/Side_effect_(computer_science)). Which makes it questionable. That's why I suggest creating a library whose only purpose is to export compiler options.

It could be an unwelcome surprise to users if they link our library and suddenly they have tons of compiler warnings they didn't ask for.

> "Modern CMake is like inheritance" That title was quite an epiphany!

Not sure I'm a fan of the title but it overall does a good job explaining things.

Here is how I always remember it.

PRIVATE is for **me** (The library)
INTERFACE is for **you** (IE The client linking the library)
PUBLIC is for **me** and **you**. (IE it's just PRIVATE + INTERFACE)

NOTE: Interface libraries can only use INTERFACE because they don't have source files to compile.

Also when building an executable just use PRIVATE. Since you don't link an executable.

PR #110, opened 2023-09-13
util: Add vk_struct_helper.hpp

### LINKED ISSUE #74: vk_typemap_helper.h
Both VVL and VEL have the same helper file:

https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/layers/vulkan/generated/vk_typemap_helper.h
https://github.com/KhronosGroup/Vulkan-ExtensionLayer/blob/main/utils/generated/vk_typemap_helper.h

charles-lunarg:
Move the vk_struct_helper.h file in Vulkan-ValidationLayers over to this repo.

Changes from that include:
 * Renaming functions
 * Placing them in the `vku` namespace
 * Adding the `InitStructHelper` class which deduces the type based on the variable that is being initialized
  * If a type isn't defined then compiler error is emitted
  
 Resolves #74


include/vulkan/utility/vk_typemap_helper.hpp
```diff
@@ -0,0 +1,1890 @@
...
+}
+
+// Find an entry of the given type in the const pNext chain
+template <typename T> const T *LvlFindInChain(const void *next) { return vku::FindStructInPNextChain<T>(next); }
+
+// Find an entry of the given type in the pNext chain
+template <typename T> T *LvlFindModInChain(void *next) { return vku::FindStructInPNextChain<T>(next); }
+
+// Find last element of pNext chain
+inline VkBaseOutStructure *LvlFindLastInChain(void *next) { return vku::FindLastStructInPNextChain(next); }
```
juan-lunarg, line 1888: 👎🏾 

If we don't fix this now when will we fix it? I understand it will be a pain to fix.

juan-lunarg (changes requested):

include/vulkan/utility/vk_typemap_helper.hpp
charles-lunarg: Thats fair, why have long term pain when some short term will lessen it.

charles-lunarg:
`vk_typemap_helper.hpp` is not a great name, since there isn't a 'typemap' in the code anywhere. `vk_struct_init.hpp` is better but I'm not sold.

juan-lunarg:
What about `vk_struct_helper.hpp`? Since this also does pNext searching?

charles-lunarg:
Thats a good idea due to the inclusion of pNext chain helpers as well.

juan-lunarg (commented):
Looks good overall aside from chromium build issue.

charles-lunarg:
build.gn has been updated.

juan-lunarg (approved):
Update PR title with correct header file

juan-lunarg:
FWIW I'd start with updating the extension layer instead of vvl.

PR #114, opened 2023-09-15
Fix warning about unreachable statement

ben-clayton:
Both sides of the `if` returned, so the `return VK_SUCCESS;` below was unreachable.


christophe-lunarg:
This issue was already resolved with https://github.com/KhronosGroup/Vulkan-Utility-Libraries/pull/113

Issue #75, opened 2023-08-23
Ensure generated code matches what scripts produce


juan-lunarg:
Currently possible to modify python scripts and not check in new generated code. Causing a mismatch.

Or vice versa modifying the C++ code which then gets overwritten by the python scripts.

PR #123, opened 2023-09-21
Remove extra semicolon after member definition

jrprice:
Some compilers complain about this.

### DIFF
--- a/include/vulkan/utility/vk_struct_helper.hpp
+++ b/include/vulkan/utility/vk_struct_helper.hpp
@@ -1852,7 +1852,7 @@ class InitStructHelper {
     void* p_next = nullptr;
   public:
     InitStructHelper() = default;
-    InitStructHelper(void *p_next) : p_next(p_next) {};
+    InitStructHelper(void *p_next) : p_next(p_next) {}
 
     template <typename T>
     operator T() {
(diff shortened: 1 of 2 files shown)


juan-lunarg (approved):

include/vulkan/utility/vk_struct_helper.hpp
```diff
@@ -1852,7 +1852,7 @@ class InitStructHelper {
     void* p_next = nullptr;
   public:
     InitStructHelper() = default;
-    InitStructHelper(void *p_next) : p_next(p_next) {};
+    InitStructHelper(void *p_next) : p_next(p_next) {}
```
juan-lunarg, line 1855: Actually you'll need to modify the codegen. Otherwise this will just end up happening again.

See struct_helper_generator.py for modifications

juan-lunarg (changes requested):

include/vulkan/utility/vk_struct_helper.hpp
juan-lunarg: It's on my list to make a CI check to catch this https://github.com/KhronosGroup/Vulkan-Utility-Libraries/issues/75

jrprice: Done, thanks.

juan-lunarg (approved):

PR #124, opened 2023-09-21
 cmake: Export library to standardize compiler/linker options 

### LINKED ISSUE #106: Export library to standardize compiler/linker options
See https://github.com/KhronosGroup/Vulkan-Utility-Libraries/issues/103 for context

This shows a clear path toward standardizing compiler / linker options across our repositories and reducing a lot of duplicate CMake code.

juan-lunarg:
Also fixes issue where compiler warnings weren't being applied to `Vulkan::LayerSettings`

closes #106

### DIFF
--- a/include/CMakeLists.txt
+++ b/include/CMakeLists.txt
@@ -16,8 +16,6 @@ set(CMAKE_FOLDER "${CMAKE_FOLDER}/VulkanUtilityHeaders")
 add_library(VulkanUtilityHeaders INTERFACE)
 add_library(Vulkan::UtilityHeaders ALIAS VulkanUtilityHeaders)
 
-lunarg_target_compiler_configurations(VulkanUtilityHeaders VUL_WERROR)
-
 # https://cmake.org/cmake/help/latest/release/3.19.html#other
 if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.19")
 	target_sources(VulkanUtilityHeaders PRIVATE
@@ -28,8 +26,6 @@ if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.19")
 	)
 endif()
 
-# NOTE: Because Vulkan::Headers header files are exposed in the public facing interface
-# we must expose this library as public to users.
 target_link_Libraries(VulkanUtilityHeaders INTERFACE Vulkan::Headers)
 
 target_include_directories(VulkanUtilityHeaders INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>)
--- a/scripts/CMakeLists.txt
+++ b/scripts/CMakeLists.txt
@@ -102,53 +102,6 @@ if (UPDATE_DEPS)
     endif()
 endif()
 
-# Configure compiler build options common to all targets
-function(lunarg_target_compiler_configurations TARGET_NAME WERROR_OPTION)
-    if (NOT ${WERROR_OPTION})
-        return()
-    endif()
-
-    if (MSVC)
-        target_compile_options(${TARGET_NAME} INTERFACE /WX)
-        target_link_options(${TARGET_NAME} INTERFACE /WX)
-    else()
-        target_compile_options(${TARGET_NAME} INTERFACE -Werror)
-    endif()
-
-    if(${CMAKE_CXX_COMPILER_ID} MATCHES "(GNU|Clang)")
-        target_compile_options(${TARGET_NAME} INTERFACE
-            -Wpedantic
-            -Wunreachable-code
-            -Wunused-function
-            -Wall
-            -Wextra
-            -Wpointer-arith
-        )
-        if(${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
-            target_compile_options(${TARGET_NAME} INTERFACE
-                -Wunreachable-code-return
-                -Wconversion
-                -Wimplicit-fallthrough
-                -Wstring-conversion
-            )
-        endif()
-    elseif(MSVC)
-        target_compile_options(${TARGET_NAME} INTERFACE
-            /W4
-            /we5038 # Enable warning about MIL ordering in constructors
-        )
-
-        # Enforce stricter ISO C++
-        target_compile_options(${TARGET_NAME} INTERFACE /permissive-)
-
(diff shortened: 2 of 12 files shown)


include/CMakeLists.txt
```diff
@@ -28,8 +26,6 @@ if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.19")
 	)
 endif()
 
-# NOTE: Because Vulkan::Headers header files are exposed in the public facing interface
-# we must expose this library as public to users.
```
juan-lunarg, line 32: Removed documentation in favor of test code where I now check INTERFACE properties.

src/CMakeLists.txt
```diff
@@ -3,4 +3,90 @@
 # Copyright 2023 LunarG, Inc.
 #
 # SPDX-License-Identifier: Apache-2.0
+
+add_library(VulkanCompilerConfiguration INTERFACE)
+add_library(Vulkan::CompilerConfiguration ALIAS VulkanCompilerConfiguration)
```
juan-lunarg, line 8: Let me know what you think of the name. I just picked something.

Linking against this library will apply the following compiler options/definitions/etc.

scripts/CMakeLists.txt
```diff
@@ -102,53 +102,6 @@ if (UPDATE_DEPS)
...
-            /we5038 # Enable warning about MIL ordering in constructors
-        )
-
-        # Enforce stricter ISO C++
-        target_compile_options(${TARGET_NAME} INTERFACE /permissive-)
-
-        target_compile_options(${TARGET_NAME} INTERFACE $<$<BOOL:${MSVC_IDE}>:/MP>) # Speed up Visual Studio builds
-
-        # Allow usage of unsafe CRT functions and minimize what Windows.h leaks
-        target_compile_definitions(${TARGET_NAME} INTERFACE _CRT_SECURE_NO_WARNINGS NOMINMAX WIN32_LEAN_AND_MEAN)
```
juan-lunarg, line 147: Note: I didn't want to export `_CRT_SECURE_NO_WARNINGS` since that should be a decision left to the authors of each library.

src/layer/CMakeLists.txt
```diff
@@ -8,8 +8,6 @@ set(CMAKE_FOLDER "${CMAKE_FOLDER}/VulkanLayerSettings")
 add_library(VulkanLayerSettings STATIC)
 add_library(Vulkan::LayerSettings ALIAS VulkanLayerSettings)
 
-lunarg_target_compiler_configurations(VulkanLayerSettings VUL_WERROR)
```
juan-lunarg, line 11: This was actually a bug. `lunarg_target_compiler_configurations` didn't work as intended.

Because it only added to the INTERFACE properties.

src/layer/CMakeLists.txt
```diff
@@ -19,6 +17,9 @@ target_sources(VulkanLayerSettings PRIVATE
...
 )
 
-# NOTE: Because Vulkan::Headers header files are exposed in the public facing interface
-# we must expose this library as public to users.
-target_link_Libraries(VulkanLayerSettings PUBLIC Vulkan::Headers)
+target_link_Libraries(VulkanLayerSettings
+   PUBLIC
+      Vulkan::Headers
+   PRIVATE
+      Vulkan::CompilerConfiguration
```
juan-lunarg, line 24: This properly adds the warnings to `VulkanLayerSettings`

christophe-lunarg (approved):
Interesting idea!

src/layer/layer_settings_manager.cpp
```diff
@@ -58,9 +58,19 @@ static std::string GetEnvironment(const char *variable) {
...
+#elif defined(_WIN32)
+    auto size = GetEnvironmentVariableA(variable, NULL, 0);
+    if (size == 0) {
+        return "";
+    }
+    char *buffer = new char[size];
+    GetEnvironmentVariableA(variable, buffer, size);
+    std::string output = buffer;
+    delete[] buffer;
+    return output;
```
juan-lunarg, line 70: Fixes `_CRT_SECURE_NO_WARNINGS`. I just took this code from VVL.

src/layer/layer_settings_manager.cpp
```diff
@@ -58,9 +58,19 @@ static std::string GetEnvironment(const char *variable) {
         result = GetAndroidProperty("debug.vulkan.screenshot");
     }
     return result;
+#elif defined(_WIN32)
```
christophe-lunarg, line 61: Do you know what compiler configuration cause these codes to generate warnings?

juan-lunarg: This one is caused by removing `_CRT_SECURE_NO_WARNINGS`

src/CMakeLists.txt
```diff
@@ -3,4 +3,90 @@
...
+    target_compile_definitions(VulkanCompilerConfiguration INTERFACE VK_USE_PLATFORM_WIN32_KHR)
+elseif(ANDROID)
+    target_compile_definitions(VulkanCompilerConfiguration INTERFACE VK_USE_PLATFORM_ANDROID_KHR)
+elseif(APPLE)
+    target_compile_definitions(VulkanCompilerConfiguration INTERFACE VK_USE_PLATFORM_METAL_EXT)
+    if (IOS)
+        target_compile_definitions(VulkanCompilerConfiguration INTERFACE VK_USE_PLATFORM_IOS_MVK)
+    else()
+        target_compile_definitions(VulkanCompilerConfiguration INTERFACE VK_USE_PLATFORM_MACOS_MVK)
+    endif()
```
charles-lunarg, line 55: I'm not actually sure this is correct, as I believe VK_USE_PLATFORM_MACOS_MVK and VK_USE_PLATFORM_IOS_MVK are deprecated. 

https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap55.html#VK_MVK_ios_surface

juan-lunarg: Ah. I didn't know. Cool beans.

juan-lunarg: Actually even though it's deprecated VVL still has code for both of these compile definitions. So I'd rather keep it.

PR #129, opened 2023-09-27
ci: Ensure generated files are up to date

### LINKED ISSUE #75: Ensure generated code matches what scripts produce

juan-lunarg:
closes #75


scripts/generate_source.py
charles-lunarg, line 1: Why does this have no changes?

.github/workflows/ci.yml
```diff
@@ -90,3 +90,16 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - name: Generate source
+        run: |
+          scripts/update_deps.py --dir external --no-build
+          scripts/generate_source.py external/Vulkan-Headers/registry/
```
charles-lunarg, line 104: This is a neat way to check for a different output, rather than having a completely different python script.

scripts/generate_source.py
juan-lunarg: chmod +x

.github/workflows/ci.yml
juan-lunarg: Yeah I was trying to come up with the laziest solution possible.

charles-lunarg (approved):

PR #131, opened 2023-09-29
Fix doc for vkuFormatIsXChromaSubsampled

juan-lunarg:
Now the documentation matches the implementation.

### DIFF
--- a/include/vulkan/utility/vk_format_utils.h
+++ b/include/vulkan/utility/vk_format_utils.h
@@ -227,7 +227,7 @@ inline bool vkuFormatIsPacked(VkFormat format);
 inline bool vkuFormatRequiresYcbcrConversion(VkFormat format);
 
 // Returns whether a VkFormat is XChromaSubsampled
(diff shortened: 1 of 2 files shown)


include/vulkan/utility/vk_format_utils.h
```diff
@@ -227,7 +227,7 @@ inline bool vkuFormatIsPacked(VkFormat format);
 inline bool vkuFormatRequiresYcbcrConversion(VkFormat format);
 
 // Returns whether a VkFormat is XChromaSubsampled
-// This corresponds to formats with _422 in their name
+// This corresponds to formats with _422 or 420 in their name
```
juan-lunarg, line 230: @spencer-lunarg does this match your original intent?

spencer-lunarg: yes,  "X Chroma Subsampled" is both, but `Y` is only 420

spencer-lunarg (approved):

charles-lunarg (approved):

PR #133, opened 2023-10-03
Add GetObjectType helper to struct_helper and Fix base_generator vendorTags logic

charles-lunarg:
This utility converst Vulkan handle types to VkObjectType enums, which is used
by the debug utils set name & tag functions.

The logic for vendorTags didn't correctly iterate through the children of tags,
causing the vendorTags list to be generated incorrectly.

### DIFF
--- a/scripts/generators/base_generator.py
+++ b/scripts/generators/base_generator.py
@@ -125,8 +125,9 @@ class BaseGenerator(OutputGenerator):
         for platform in self.registry.tree.findall('platforms/platform'):
             self.vk.platforms[platform.get('name')] = platform.get('protect')
 
-        for tag in self.registry.tree.findall('tags'):
-            self.vk.vendorTags.append(tag.get('name'))
+        for tags in self.registry.tree.findall('tags'):
+            for tag in tags.findall('tag'):
+                self.vk.vendorTags.append(tag.get('name'))
 
         # No way known to get this from the XML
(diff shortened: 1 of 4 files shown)


scripts/generators/base_generator.py
```diff
@@ -125,8 +125,9 @@ def beginFile(self, genOpts):
         for platform in self.registry.tree.findall('platforms/platform'):
             self.vk.platforms[platform.get('name')] = platform.get('protect')
 
-        for tag in self.registry.tree.findall('tags'):
-            self.vk.vendorTags.append(tag.get('name'))
+        for tags in self.registry.tree.findall('tags'):
+            for tag in tags.findall('tag'):
```
spencer-lunarg, line 129: clearly this was not being used before... while we have this small time of divergence (while I get this file in the Vulkan-Headers), can you apply this fix in the VVL as well :pray:

spencer-lunarg (approved):

scripts/generators/base_generator.py
charles-lunarg: You could say,
*puts on sunglasses*
I already did.

charles-lunarg: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/6611

juan-lunarg (approved):

PR #137, opened 2023-10-04
Make GetObjectType specializations inline

mikes-lunarg:
Avoid multiple definition compiler error


charles-lunarg (approved):
Whoops, thanks for finding and fixing this rather obvious mistake!

juan-lunarg (approved):

PR #139, opened 2023-10-05
util: Fix vku::GetObjectType() broken for 32 bit builds

charles-lunarg:
Non Dispatchable handles are not typed in 32 bit builds, so the vku::GetObjectHandle() helper fails to compile. The solution is to just not define the helper in 32 bit builds.

### DIFF
--- a/include/vulkan/utility/vk_struct_helper.hpp
+++ b/include/vulkan/utility/vk_struct_helper.hpp
@@ -1175,6 +1175,7 @@ template<typename T> VkObjectType GetObjectType() {
     return VK_OBJECT_TYPE_UNKNOWN;
 }
 
+#if VK_USE_64_BIT_PTR_DEFINES == 1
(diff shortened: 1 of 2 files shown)


include/vulkan/utility/vk_struct_helper.hpp
```diff
@@ -1226,6 +1227,7 @@ template<> inline VkObjectType GetObjectType<VkMicromapEXT>() { return VK_OBJECT
 template<> inline VkObjectType GetObjectType<VkOpticalFlowSessionNV>() { return VK_OBJECT_TYPE_OPTICAL_FLOW_SESSION_NV; }
 template<> inline VkObjectType GetObjectType<VkShaderEXT>() { return VK_OBJECT_TYPE_SHADER_EXT; }
 
+#endif VK_USE_64_BIT_PTR_DEFINES == 1
```
juan-lunarg, line 1230:
```suggestion
#endif // VK_USE_64_BIT_PTR_DEFINES == 1
```

juan-lunarg (approved):

PR #141, opened 2023-10-05
cmake: Only create 1 test executable

juan-lunarg:
Simplifies the code and makes adding Android testing much easier.

Since you only have to deploy 1 binary.


tests/layer/CMakeLists.txt
```diff
@@ -3,130 +3,16 @@
...
-add_executable(test_layer_setting_file)
-
-target_include_directories(test_layer_setting_file PRIVATE
-    ${CMAKE_SOURCE_DIR}/src/layer
-)
-
-target_sources(test_layer_setting_file PRIVATE
     test_setting_file.cpp
+    test_setting_util.cpp
 )
```
charles-lunarg, line 18: Why even have this file? Seems fine to just put everything in tests/CMakeLists.txt since thats how the other libraries are doing it.

juan-lunarg: Sure I don't mind.

juan-lunarg: 👍🏾

charles-lunarg (approved):

PR #149, opened 2023-10-13
ci: Add dependabot

juan-lunarg:
Ensures actions are kept up to date

Update ci.yml to not create duplicate workflow when dependabot creates branches on the repo

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,10 +7,12 @@
 name: ci
 
 on:
-    push:
-    pull_request:
-        branches:
-            - main
+  push:
+    branches:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -7,10 +7,12 @@
...
-    push:
-    pull_request:
-        branches:
-            - main
+  push:
+    branches:
+    - main
+  pull_request:
+    branches:
+    - main
```
juan-lunarg, line 15: This prevents the situation where GHA runs twice because the user didn't use a fork. Which is helpful since dependabot doesn't use forks.

Issue #73, opened 2023-08-23
lvt_function_pointers & vk_function_pointers


juan-lunarg:
Both VVL and VEL use very similar generated code:

VVL
- https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/layers/vulkan/generated/vk_function_pointers.h
- https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/layers/vulkan/generated/vk_function_pointers.cpp

VEL
- https://github.com/KhronosGroup/Vulkan-ExtensionLayer/blob/main/utils/generated/lvt_function_pointers.h
- https://github.com/KhronosGroup/Vulkan-ExtensionLayer/blob/main/utils/generated/lvt_function_pointers.cpp

juan-lunarg:
> Opened this up a month ago, but the reality is that the solution is to use volk, not write another new library. If there is future interest in library for function loading in this repo, it can be revisited.

PR #152, opened 2023-10-16
Add vk_function_loader.hpp

charles-lunarg:
Adds a new header only utility library which contains a loader for the Vulkan-Loader,
tables to hold vulkan function pointers, and init functions to fill them out.

This is for #73 - will close the issue when VVL & VEL are using this library.

### DIFF
--- a/include/CMakeLists.txt
+++ b/include/CMakeLists.txt
@@ -23,10 +23,11 @@ if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.19")
 		vulkan/vk_enum_string_helper.h
 		vulkan/utility/vk_format_utils.h
 		vulkan/utility/vk_struct_helper.hpp
+		vulkan/utility/experimental/vk_function_loader.hpp
 	)
 endif()
 
-target_link_Libraries(VulkanUtilityHeaders INTERFACE Vulkan::Headers)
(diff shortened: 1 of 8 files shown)


tests/test_setting_env.cpp
```diff
@@ -439,4 +439,6 @@ TEST(test_layer_setting_env, vkuGetLayerSettingValues_String) {
     EXPECT_EQ(2, value_count);
 
     vkuDestroyLayerSettingSet(layerSettingSet, nullptr);
+
+    SetEnv("VK_LUNARG_TEST_MY_SETTING=");
```
juan-lunarg, line 443: Make a separate PR so we can just merge this fix in.

juan-lunarg (changes requested):
I'd like to see a branch with the functionality implemented in VEL or VVL.

tests/test_setting_env.cpp
charles-lunarg: Done
(doing some more changes before pushing).

charles-lunarg:
Note that this PR relies on #155 so I will rebase when that is merged.

charles-lunarg:
Opened this up a month ago, but the reality is that the solution is to use volk, not write another new library. If there is future interest in library for function loading in this repo, it can be revisited.

PR #154, opened 2023-10-17
Unset env-var in layer tests

charles-lunarg:
Need to unset an env-var in order to allow the vul_tests executable to pass all tests. This is because environment variables leak between tests when not running with CTest.

### DIFF
--- a/tests/test_setting_env.cpp
+++ b/tests/test_setting_env.cpp
@@ -439,4 +439,6 @@ TEST(test_layer_setting_env, vkuGetLayerSettingValues_String) {
     EXPECT_EQ(2, value_count);
 
     vkuDestroyLayerSettingSet(layerSettingSet, nullptr);
+
+    SetEnv("VK_LUNARG_TEST_MY_SETTING=");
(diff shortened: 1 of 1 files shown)


juan-lunarg (approved):

tests/test_setting_env.cpp
```diff
@@ -439,4 +439,6 @@ TEST(test_layer_setting_env, vkuGetLayerSettingValues_String) {
     EXPECT_EQ(2, value_count);
 
     vkuDestroyLayerSettingSet(layerSettingSet, nullptr);
+
+    SetEnv("VK_LUNARG_TEST_MY_SETTING=");
```
juan-lunarg, line 443: FFTI

This may be a case where a class for environment variables may be merited. IE the destructor would unset the environment vairable.

charles-lunarg: If a lot of tests are written using env-vars, definitely worth it. Since this is the only test so far, I'm happy to leave it for now.

PR #155, opened 2023-10-17
Remove consecutive platform defines

charles-lunarg:
Generated code must macro-guard platform specific code, but did it in a naive fashion where consecutive guards for the same platform repeated the macro. With the help of PlatformGuardHelper, the code generation will elide redundant macro guards.


juan-lunarg (approved):
LGTM

FYI @spencer-lunarg

scripts/generators/dispatch_table_generator.py
```diff
@@ -8,6 +8,7 @@
 
 import os
 from generators.base_generator import BaseGenerator
```
spencer-lunarg, line 10: nit: 

```suggestion
from generators.base_generator import BaseGenerator, PlatformGuardHelper
```

scripts/generators/base_generator.py
```diff
@@ -70,6 +70,20 @@ def SetMergedApiNames(names: str) -> None:
     global mergedApiNames
     mergedApiNames = names
 
+# Used to elide platform guards together, so redundant #endif then #ifdefs are removed
+class PlatformGuardHelper():
+    def __init__(self):
+        self.current_guard = None
```
spencer-lunarg, line 76: not sure if this is the "python way" but the rest of the variables are camelCase so for sake of being consistent

`self.currentGuard`

spencer-lunarg (changes requested):
not disagreeing with this change, I think its a super good thing and cleans up something I hated

What I am not sure of is `base_generator.py` should start to become a util bin

it solo job I think should be to create `VulkanObject`... I feel this change in the future this would lead to just everyone wanting to add their non-related utils in this file

charles-lunarg:
Sure, I can agree that it’s not great to have a dumping ground for utils. I’m not sure of a better place though, so I’m open to suggestions.

spencer-lunarg:
the VVL has a `generator_utils.py`, wouldn't be the worse for the short term to move it to its own file, curious to see how many "utils" will actually be shared

scripts/generators/base_generator.py
charles-lunarg: The python linter extension I have installed in vscode tells me to use snake_case, but I'm happy to use camelCase. I am not sure who or what tells the linter which style to use, so it may just be a default that isn't applicable to us (as in I haven't configured the extension).

spencer-lunarg: if we are moving into its own utils, feel free to start snake_casing

spencer-lunarg (approved):

charles-lunarg:
I'll use camelCase because the rest of the project appears to use it. Easier to keep in line with everything else and if someone wants to change it in the future, they are more than welcome to.

PR #158, opened 2023-11-06
Remove unnecessary std filesystem

ziga-lunarg:
Some users of VUL don't have support for std::filesystem, and there is no real reason that std::filesystem is used here


juan-lunarg:
> Some users of VUL don't have support for std::filesystem, and there is no real reason that std::filesystem is used here

Who are the users? VUL will not compile without a C++17 compiler.

juan-lunarg (dismissed):

juan-lunarg:
FYI @christophe-lunarg / @mikes-lunarg

christophe-lunarg:
Erm, yes this doesn't seem right to me, specifically "VUL will not compile without a C++17 compiler." unless if the idea is to support older compilers for specific reasons that are not mentioned here.

PR #159, opened 2023-11-10
build: Update to header 1.3.270

mikes-lunarg:
- Update known-good
- Generate source


juan-lunarg (approved):

juan-lunarg:
Love tiny updates like this.

PR #160, opened 2023-11-10
Don't install VulkanCompilerConfiguration

juan-lunarg:
This isn't used anywhere at the moment.

The idea was for projects to link against this fake library.

But it just does too much at the moment.

Example(s):

Warnings are too high for all projects
Not all projects need/want WSI support

Instead of leaving it around to eventually cause a problem remove it.


juan-lunarg:
closing forgot this breaks installation due to installation of static library

PR #163, opened 2023-11-17
Fix default env setting prefix for Android

kocdemir:
Vulkan-Profiles layer (and possibly other users of the library) is not working correctly on Android, due to multiple point symbols added for the environment variable checks. This CL removes the unnecessary '.' character to fix the issue.


juan-lunarg:
@christophe-lunarg ideally we add a test to prevent a regression. Should the test be added here, profiles, or both?

christophe-lunarg (approved):

PR #167, opened 2023-12-01
build: Update to header 1.3.272

mikes-lunarg:
- Update known-good
- Generate source
- Remove vk_layer_settings_ext.h
- Change all usage of VkLayerSettingEXT::count to use the new name valueCount from the public header.


juan-lunarg:
LGTM. @christophe-lunarg should give the approval though.

juan-lunarg:
Likely worth updating the layer_configuration.md and making sure we highlight this in the SDK release notes.

christophe-lunarg (approved):

PR #170, opened 2023-12-07
ci: Add Windows ARM to CI

juan-lunarg:


.github/workflows/ci.yml
```diff
@@ -34,6 +34,19 @@ jobs:
...
 
+  windows-arm64:
+    runs-on: windows-latest
+    env:
+      CMAKE_GENERATOR: Ninja
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ilammy/msvc-dev-cmd@v1
+        with:
+          arch: amd64_arm64
```
juan-lunarg, line 45: This ensures we are cross-compiling for arm64

charles-lunarg (commented):
So is this more a 'cross compilation' build rather than a build on arm computers?

juan-lunarg:
> So is this more a 'cross compilation' build rather than a build on arm computers?

Yep.

PR #171, opened 2023-12-08
build: Make find_package logic more robust

aqnuep:
This fixes the problem whereas in the generated `VulkanUtilityLibrariesConfig.cmake` the value of `PACKAGE_PREFIX_DIR` may get overwritten by the `VulkanHeaders` discovery process.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,11 +63,11 @@ if (VUL_IS_TOP_LEVEL)
     # Create VulkanUtilityLibrariesConfig.cmake
     set(VUL_EXPORT_TARGETS ${CMAKE_INSTALL_LIBDIR}/cmake/VulkanUtilityLibraries/VulkanUtilityLibraries-targets.cmake)
     file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/VulkanUtilityLibrariesConfig.cmake.in" [=[
-        @PACKAGE_INIT@
-
         include(CMakeFindDependencyMacro)
         # NOTE: Because VulkanHeaders is a PUBLIC dependency it needs to be found prior to VulkanUtilityLibraries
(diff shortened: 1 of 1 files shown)


juan-lunarg:
Hmm. Looking at VulkanSC-Headers CMakeLists.txt it's quite different from Vulkan-Headers. That's likely the discrepancy.

Vulkan-Headers doesn't use `configure_package_config_file`.

aqnuep:
> Hmm. Looking at VulkanSC-Headers CMakeLists.txt it's quite different from Vulkan-Headers. That's likely the discrepancy.

It's likely based on an earlier revision of Vulkan-Headers.

Does that mean this PR is not going to be accepted and instead we need to update VulkanSC-Headers?

aqnuep:
FWIW, my argument for accepting this nonetheless is that IMO that Vulkan-Headers should be treated as a black-box, and the package scripts should not rely on the internal implementation of it.

juan-lunarg:
> FWIW, my argument for accepting this nonetheless is that IMO that Vulkan-Headers should be treated as a black-box, and the package scripts should not rely on the internal implementation of it.

I agree. It should be merged.

juan-lunarg (approved):

juan-lunarg:
> Does that mean this PR is not going to be accepted and instead we need to update VulkanSC-Headers?

I'd advise updating VulkanSC-Headers regardless of this PR being merged.

juan-lunarg:
Could you rename the commit / PR name though. Something like

`Make find_package logic more robust`

Since this PR doesn't fix anything for us.

aqnuep:
Sure, and thank you, @juan-lunarg. I agree that it's worth doing an update to the VulkanSC-Headers build system as it's quite out of date.

PR #187, opened 2024-03-19
Move vk_safe_struct.h to VUL

jeremyg-lunarg:
This code was being generated in both Vulkan-ValidationLayers and Vulkan-ExtensionLayer. Further uses are on the horizon so lets stop the copypasta.


jeremyg-lunarg:
This is draft until I prove it works for VUL and VEL

include/vulkan/utility/concurrent_unordered_map.h
```diff
@@ -0,0 +1,212 @@
...
+            result |= maps[h].empty();
+        }
+        return result;
+    }
+
+  private:
+    static const int BUCKETS = (1 << BUCKETSLOG2);
+
+    // TODO: template magic to allow other unordered_maps to be used here
+    std::unordered_map<Key, T, Hash> maps[BUCKETS];
```
jeremyg-lunarg, line 198: This TODO is where we'd want to use  the robin hood or unordered_dense hashmaps via the vvl::unordered_map alias in the original code.

scripts/common_ci.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/python3 -i
```
jeremyg-lunarg, line 1: I pulled over lots of python changes to make the safe struct generators and clang-format work. It is ugly copy copy though.

jeremyg-lunarg (commented):
This would be a good opportunity to change the safe names from safe_VkInstanceCreateInfo to something like `vku::safe::InstanceCreateInfo` if we want to do that.

src/vulkan/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+    vk_safe_struct_vendor.cpp
+    vk_safe_struct_manual.cpp
+)
+
+target_link_Libraries(VulkanUtilityHeaders 
+   PUBLIC
+      Vulkan::Headers
+   PRIVATE
+      Vulkan::CompilerConfiguration
+)
```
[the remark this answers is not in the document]
jeremyg-lunarg: pick which confusion you prefer. :)

scripts/generators/generator_utils.py
```diff
@@ -1,25 +1,157 @@
...
+        if extra_newline:
+            out.append('\n')
+        if self.current_guard is not guard and guard is not None:
+            out.append(f'#ifdef {guard}\n')
+        self.current_guard = guard
         return out
+
+# The SPIR-V grammar json doesn't have an easy way to detect these, so have listed by hand
+# If we are missing one, its not critical, the goal of this list is to reduce the generated output size
+def IsNonVulkanSprivCapability(capability):
```
spencer-lunarg, line 91: not needed

scripts/generators/vulkan_object.py
```diff
@@ -1,10 +1,19 @@
 #!/usr/bin/python3 -i
```
spencer-lunarg, line 1: so sorry I haven't got to a point where this is all unified in put the `Vulkan-Headers` yet, but we can have a seperate PR where we unify `vulkan_object.py` and `base_generator.py` across everywhere

scripts/generators/generator_utils.py
jeremyg-lunarg: I'm trying to keep the customization of scripts to a minimum and just using the VVL versions as a the known good version

jeremyg-lunarg:
> This would be a good opportunity to change the safe names from safe_VkInstanceCreateInfo to something like `vku::safe::InstanceCreateInfo` if we want to do that.

Changing to this naming convention is problematic because safe_VkMemoryBarrier becomes vku::safe::MemoryBarrier. Windows has a macro named MemoryBarrier which then breaks all over the place. Both this code and client code would need to #undef that macro.

scripts/generators/generator_utils.py
```diff
@@ -1,25 +1,148 @@
...
-            if guard != None:
-                out.append(f'#ifdef {guard}\n')
-            self.currentGuard = guard
+        if self.current_guard is not guard and self.current_guard is not None:
+            out.append(f'#endif  // {self.current_guard}\n')
+        if extra_newline:
+            out.append('\n')
+        if self.current_guard is not guard and guard is not None:
+            out.append(f'#ifdef {guard}\n')
+        self.current_guard = guard
```
charles-lunarg, line 77:
```suggestion
        if self.current_guard != guard and self.current_guard is not None:
            out.append(f'#endif  // {self.current_guard}\n')
        if extra_newline:
            out.append('\n')
        if self.current_guard != guard and guard is not None:
            out.append(f'#ifdef {guard}\n')
        self.current_guard = guard
```
Apparently replacing `!=` with `is not` provokes different behavior.

include/vulkan/vk_enum_string_helper.h
```diff
@@ -903,12 +903,20 @@ static inline const char* string_VkStructureType(VkStructureType input_value) {
...
+#endif  // VK_ENABLE_BETA_EXTENSIONS
+#ifdef VK_ENABLE_BETA_EXTENSIONS
         case VK_STRUCTURE_TYPE_EXECUTION_GRAPH_PIPELINE_SCRATCH_SIZE_AMDX:
             return "VK_STRUCTURE_TYPE_EXECUTION_GRAPH_PIPELINE_SCRATCH_SIZE_AMDX";
+#endif  // VK_ENABLE_BETA_EXTENSIONS
+#ifdef VK_ENABLE_BETA_EXTENSIONS
         case VK_STRUCTURE_TYPE_EXECUTION_GRAPH_PIPELINE_CREATE_INFO_AMDX:
             return "VK_STRUCTURE_TYPE_EXECUTION_GRAPH_PIPELINE_CREATE_INFO_AMDX";
+#endif  // VK_ENABLE_BETA_EXTENSIONS
+#ifdef VK_ENABLE_BETA_EXTENSIONS
```
charles-lunarg, line 919:
```suggestion
        case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_ENQUEUE_PROPERTIES_AMDX:
            return "VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_ENQUEUE_PROPERTIES_AMDX";
        case VK_STRUCTURE_TYPE_EXECUTION_GRAPH_PIPELINE_SCRATCH_SIZE_AMDX:
            return "VK_STRUCTURE_TYPE_EXECUTION_GRAPH_PIPELINE_SCRATCH_SIZE_AMDX";
        case VK_STRUCTURE_TYPE_EXECUTION_GRAPH_PIPELINE_CREATE_INFO_AMDX:
            return "VK_STRUCTURE_TYPE_EXECUTION_GRAPH_PIPELINE_CREATE_INFO_AMDX";
```

charles-lunarg: This is what ticked me off to the guard_helper regressing.

charles-lunarg (changes requested):

scripts/generators/generator_utils.py
jeremyg-lunarg: ok. that means VVL is broken too, because I just copied the 'newer' scripts over... We have script madness.

jeremyg-lunarg: should be fixed now. I'll push a fix to VVL in a sec

jeremyg-lunarg: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7787

src/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ elseif(MSVC)
     target_compile_options(VulkanCompilerConfiguration INTERFACE
         /W4
         /we5038 # Enable warning about MIL ordering in constructors
+	/wd4324 # Disable warning about alignment padding
```
charles-lunarg, line 31: Normally I'd be hesitant to add warnings as this target is intended to be consumed downstream. HOWEVER 1. I don't think anyone uses it and 2. this w4324 seems to be a real nonsense warning due to alignment.

charles-lunarg (approved):
Most of the changes are not code I really *should* review. As a lot of this code comes straight from VVL. But I did go over the bits that aren't in VVL well enough to feel confident.

Of course, it would be great if @jeremyg-lunarg used this PR in VVL before merging it to make sure it 'works' but safe structs is SO HUGE that I don't see a reason to hold up this PR while VVL is moved over. If fixes are needed to VUL's safe structs, that is just gonna have to happen.

jeremyg-lunarg:
> Of course, it would be great if @jeremyg-lunarg used this PR in VVL before merging

There are draft PRs for VEL (https://github.com/KhronosGroup/Vulkan-ExtensionLayer/pull/378)and VVL (https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7717)  that use this PR.

src/CMakeLists.txt
jeremyg-lunarg: this warning comes from vku::concurrent::unordered_map which purposefully adds padding for better multithreaded performance.

PR #201, opened 2024-04-20
Revert "safestruct: Remove safe_VkDescriptorDataEXT"

ziga-lunarg:
Revert #196

VkDescriptorDataEXT needs to be safe.

When running with validation layers the CTS test `dEQP-VK.binding_model.descriptor_buffer.single.graphics_vert_sampled_image` and other descriptor buffer tests crash after the tests finish. 

If `VkDescriptorDataEXT` is not safe, then in `layer_chassis_dispatch_manual.cpp` in function `DispatchGetDescriptorEXT()` the descriptors are overwritten with the unwrapped handles, because `local_pDescriptorInfo.data` has the same pointers as `pDescriptorInfo->data`, and when tests finish the invalid unwrapped handles are used to delete the descriptors (sampler is this case), which leads to:
```
Test case 'dEQP-VK.binding_model.descriptor_buffer.single.graphics_vert_sampled_image'..
VUID-vkDestroySampler-sampler-parameter(ERROR / SPEC): msgNum: -576740016 - Validation Error: [ VUID-vkDestroySampler-sampler-parameter ] Object 0: handle = 0x27f96f1bcd0, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0xdd9fa550 | vkDestroySampler(): sampler Invalid VkSampler Object 0x27ff7abb6b0. The Vulkan spec states: If sampler is not VK_NULL_HANDLE, sampler must be a valid VkSampler handle (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkDestroySampler-sampler-parameter)
    Objects: 1
        [0] 0x27f96f1bcd0, type: 1, name: NULL
Assertion failed: object_table.contains(object), file C:\Projects\Vulkan-ValidationLayers\layers\thread_tracker/thread_safety_validation.h, line 116
```

### DIFF
--- a/include/vulkan/utility/vk_safe_struct.hpp
+++ b/include/vulkan/utility/vk_safe_struct.hpp
@@ -13494,11 +13494,33 @@ struct safe_VkDescriptorBufferBindingPushDescriptorBufferHandleEXT {
         return reinterpret_cast<VkDescriptorBufferBindingPushDescriptorBufferHandleEXT const*>(this);
     }
 };
+union safe_VkDescriptorDataEXT {
+    const VkSampler* pSampler{};
+    const VkDescriptorImageInfo* pCombinedImageSampler;
+    const VkDescriptorImageInfo* pInputAttachmentImage;
+    const VkDescriptorImageInfo* pSampledImage;
+    const VkDescriptorImageInfo* pStorageImage;
+    safe_VkDescriptorAddressInfoEXT* pUniformTexelBuffer;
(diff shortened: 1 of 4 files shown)


charles-lunarg (approved):

spencer-lunarg (approved):
something is wrong here as still not sure why this is needed... will take a look later, happy to un-crash CTS for now

PR #202, opened 2024-04-24
safestruct: Remove safe_VkDescriptorDataEXT

spencer-lunarg:
version 2 of https://github.com/KhronosGroup/Vulkan-Utility-Libraries/pull/196 

This was removed in https://github.com/KhronosGroup/Vulkan-Utility-Libraries/pull/201

but then fixed in https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7898

and this time it is ready to go


charles-lunarg (approved):
At first I was confused cause the Pr is like "remove the reverted removal of safe_struct" but then I read the changes and everything made sense.

Still, the reality of 'making it not break right now' taking priority over 'making the correct fix later' leads to some fun change descriptions.

ziga-lunarg (approved):

PR #203, opened 2024-04-25
iwyu: missing #include <memory> for std::unique_ptr

selfisekai:


charles-lunarg (approved):

charles-lunarg:
Ugh it looks like arm64 (aka macOS) only supports python 3.10 and up? Only positive note is that it appears macOS CI is running on arm computers now, instead of being intel chips only.

PR #218, opened 2024-06-10
safe_struct: Make FreePnextChain() non-recursive

jeremyg-lunarg:


charles-lunarg (commented):
Looks appropriate - not sure how to validate this manually. Best thing is to have it pass an ASAN run in VVL so that it isn't leaking memory.

charles-lunarg (approved):
Honestly this is a more procedural change than I thought (pun intended). I'm going to go ahead and approve to unblock you - I don't have the tools or time to validate this myself and would ask that you give your best judgement before merging. The ASAN test is nice, but may be hard to do without a local build already setup. Regardless, earnestly appreciate going out and fixing this at the root rather than working around it.

spencer-lunarg (approved):
Confirms this solves https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8110 and doesn't create a huge stack

jeremyg-lunarg:
> Confirms this solves [KhronosGroup/Vulkan-ValidationLayers#8110](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8110) and doesn't create a huge stack

Thanks!

PR #220, opened 2024-06-24
build: Do not specialize source tree for API variants

aqnuep:
In order to avoid having to deal with mixed generated vs manual sources / headers in this repo, we decided to not maintain separate source trees at all for API variants downstream, so this change removes the existing path parameterizations based on the target API variant.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,7 +41,7 @@ if (VUL_IS_TOP_LEVEL)
     include(CMakePackageConfigHelpers)
 
     install(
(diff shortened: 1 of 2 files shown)


spencer-lunarg (approved):
cc @charles-lunarg

charles-lunarg:
@aqnuep Is this ready to merge? (I'm unsure if you have write access so want to double check that you aren't delaying merging for some reason)

aqnuep:
Yes, it should be ready, I was only waiting for your approval.

PR #222, opened 2024-06-27
Better declare GetObjectType()'s lack of 32 bit support

charles-lunarg:
Create a new default specialization for 32 bit which indicates to the
user that the GetObjectType() helper does not work in 32 bit builds.


artem-lunarg (approved):
LGTM. Thanks for the quick fix.

PR #223, opened 2024-06-28
build: Update to header 1.3.289

mikes-lunarg:


ziga-lunarg (approved):

mikes-lunarg:
@charles-lunarg Looks like I need to do something similar to this: https://github.com/KhronosGroup/Vulkan-Headers/commit/e3c37e6e184a232e10b01dff5a065ce48c047f88

charles-lunarg:
> @charles-lunarg Looks like I need to do something similar to this: [KhronosGroup/Vulkan-Headers@e3c37e6](https://github.com/KhronosGroup/Vulkan-Headers/commit/e3c37e6e184a232e10b01dff5a065ce48c047f88)

You can just set `-DVULKAN_HEADERS_ENABLE_MODULE=OFF` for the time being.

charles-lunarg:
If its causing CI to fail across the board, maybe updating Vulkan-Headers to set module support to OFF for the time being is the better option?

mikes-lunarg:
> If its causing CI to fail across the board, maybe updating Vulkan-Headers to set module support to OFF for the time being is the better option?

It is easy enough to opt out of modules for a while, modifying the upstream defaults is more work at this point

charles-lunarg:
Sounds good. Frankly, its just annoying that MSVC *has* fixed the bug but its not in the runners just yet.

charles-lunarg (approved):

PR #231, opened 2024-08-30
build: Update to header 1.3.295

mikes-lunarg:


src/vulkan/vk_safe_struct_utils.cpp
```diff
@@ -60,12 +60,6 @@ void *SafePnextCopy(const void *pNext, PNextCopyState* copy_state) {
             case VK_STRUCTURE_TYPE_SHADER_MODULE_CREATE_INFO:
                 safe_pNext = new safe_VkShaderModuleCreateInfo(reinterpret_cast<const VkShaderModuleCreateInfo *>(pNext), copy_state, false);
                 break;
-            case VK_STRUCTURE_TYPE_COMPUTE_PIPELINE_CREATE_INFO:
-                safe_pNext = new safe_VkComputePipelineCreateInfo(reinterpret_cast<const VkComputePipelineCreateInfo *>(pNext), copy_state, false);
-                break;
-            case VK_STRUCTURE_TYPE_GRAPHICS_PIPELINE_CREATE_INFO:
-                safe_pNext = new safe_VkGraphicsPipelineCreateInfo(reinterpret_cast<const VkGraphicsPipelineCreateInfo *>(pNext), copy_state, false);
-                break;
```
charles-lunarg, line 68: Why are these removed?

src/vulkan/vk_safe_struct_utils.cpp
```diff
@@ -937,9 +937,6 @@ void *SafePnextCopy(const void *pNext, PNextCopyState* copy_state) {
             case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_ENQUEUE_PROPERTIES_AMDX:
                 safe_pNext = new safe_VkPhysicalDeviceShaderEnqueuePropertiesAMDX(reinterpret_cast<const VkPhysicalDeviceShaderEnqueuePropertiesAMDX *>(pNext), copy_state, false);
                 break;
-            case VK_STRUCTURE_TYPE_EXECUTION_GRAPH_PIPELINE_CREATE_INFO_AMDX:
-                safe_pNext = new safe_VkExecutionGraphPipelineCreateInfoAMDX(reinterpret_cast<const VkExecutionGraphPipelineCreateInfoAMDX *>(pNext), copy_state, false);
-                break;
```
charles-lunarg, line 942: Also this.

src/vulkan/vk_safe_struct_utils.cpp
```diff
@@ -1007,9 +1004,6 @@ void *SafePnextCopy(const void *pNext, PNextCopyState* copy_state) {
             case VK_STRUCTURE_TYPE_PIPELINE_VIEWPORT_COARSE_SAMPLE_ORDER_STATE_CREATE_INFO_NV:
                 safe_pNext = new safe_VkPipelineViewportCoarseSampleOrderStateCreateInfoNV(reinterpret_cast<const VkPipelineViewportCoarseSampleOrderStateCreateInfoNV *>(pNext), copy_state, false);
                 break;
-            case VK_STRUCTURE_TYPE_RAY_TRACING_PIPELINE_CREATE_INFO_NV:
-                safe_pNext = new safe_VkRayTracingPipelineCreateInfoNV(reinterpret_cast<const VkRayTracingPipelineCreateInfoNV *>(pNext), copy_state, false);
-                break;
```
charles-lunarg, line 1012: And this.

src/vulkan/vk_safe_struct_utils.cpp
```diff
@@ -1858,9 +1849,6 @@ void *SafePnextCopy(const void *pNext, PNextCopyState* copy_state) {
             case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_ACCELERATION_STRUCTURE_PROPERTIES_KHR:
                 safe_pNext = new safe_VkPhysicalDeviceAccelerationStructurePropertiesKHR(reinterpret_cast<const VkPhysicalDeviceAccelerationStructurePropertiesKHR *>(pNext), copy_state, false);
                 break;
-            case VK_STRUCTURE_TYPE_RAY_TRACING_PIPELINE_CREATE_INFO_KHR:
-                safe_pNext = new safe_VkRayTracingPipelineCreateInfoKHR(reinterpret_cast<const VkRayTracingPipelineCreateInfoKHR *>(pNext), copy_state, false);
-                break;
```
charles-lunarg, line 1863: This too.

src/vulkan/vk_safe_struct_utils.cpp
```diff
@@ -1925,12 +1913,6 @@ void FreePnextChain(const void *pNext) {
         case VK_STRUCTURE_TYPE_SHADER_MODULE_CREATE_INFO:
             delete reinterpret_cast<safe_VkShaderModuleCreateInfo *>(header);
             break;
-        case VK_STRUCTURE_TYPE_COMPUTE_PIPELINE_CREATE_INFO:
-            delete reinterpret_cast<safe_VkComputePipelineCreateInfo *>(header);
-            break;
-        case VK_STRUCTURE_TYPE_GRAPHICS_PIPELINE_CREATE_INFO:
-            delete reinterpret_cast<safe_VkGraphicsPipelineCreateInfo *>(header);
-            break;
```
charles-lunarg, line 1933: As well as this - And I'm getting tired of commenting on each change so I wont.

charles-lunarg (commented):
Seems like the code gen removed a bunch of structs from the safe structs.

mikes-lunarg:
> Seems like the code gen removed a bunch of structs from the safe structs.

It is a bit odd, I'm looking at it now. When safe struct was part of VVL I know it only made deep copies for structs with embedded handle ids that needed to be remapped. If the header update changed the "structextends" dependency tree to remove some possible pNext structs with handles, that might have caused the codegen to no longer care about them.

charles-lunarg (approved):
I'll go ahead and approve so you can move forward with the header update. Seems that something in the xml changed, and while it may not be a big deal it definitely was something unexpected.

mikes-lunarg:
@charles-lunarg Turns out this was a good catch!

The reason those were removed was because the various pipeline create info structs no longer  "structextend" VkPipelineCreateInfoKHR in the xml. It turns out that they don't behave like true extension structs and the adding them as such screws up other parts of the spec build (like determining the base struct of pNext chain)

_but..._

Since they are still valid in the pNext chain, I added a manual workaround to include them back

charles-lunarg:
Sounds like we understand the root cause then! I already reviewed it so I can't approve it again. So consider this a second approval

PR #235, opened 2024-09-20
tests: Fix windows warnings

ziga-lunarg:
With latest visual studio, there are a lot of new warnings about comparing mismatched types


charles-lunarg (approved):
I'm sure in some way this is really a googletest issue, but its much easier to just get the issue resolved here than try to find & fix it upstream.

PR #236, opened 2024-09-26
build: Update to header 1.3.296

spencer-lunarg:
My attempt to update the headers here @mikes-lunarg (so I can target my VVL changes at least somewhere public for short term (aka my fork)) ... if this is good, happy to merge this also

### DIFF
--- a/scripts/generators/safe_struct_generator.py
+++ b/scripts/generators/safe_struct_generator.py
@@ -41,8 +41,6 @@ class SafeStructOutputGenerator(BaseGenerator):
             'VkMicromapBuildInfoEXT',
             'VkAccelerationStructureTrianglesOpacityMicromapEXT',
             'VkAccelerationStructureTrianglesDisplacementMicromapNV',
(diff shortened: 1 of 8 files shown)


scripts/generators/safe_struct_generator.py
```diff
@@ -41,8 +41,6 @@ def __init__(self):
             'VkMicromapBuildInfoEXT',
             'VkAccelerationStructureTrianglesOpacityMicromapEXT',
             'VkAccelerationStructureTrianglesDisplacementMicromapNV',
-            # The VkDescriptorType field needs to handle every type which is something best done manually
-            'VkDescriptorDataEXT',
```
spencer-lunarg, line 45: I originally listed them here, doubled checked now, we don't do this manually and instead can auto generate it with `self.union_of_pointers = [ ]`

ziga-lunarg (approved):

charles-lunarg (approved):

PR #237, opened 2024-09-27
Fix usage of __system_property_read_callback for Android<26

vrepetsky:
Modify the function for getting system properties on Android: 
*Use the new __system_property_read_callback API that appeared in Android O / API level 26 when available. It is the preferred way because it allows handling long property names and better property caching. 
*When not available use the deprecated __system_property_get function.


spencer-lunarg (approved):
@vrepetsky you will need to fix the clang-formatting issue to make CI happy

PR #244, opened 2024-10-30
CMake: Added `VUL_ENABLE_INSTALL` to top level cmake file.

mokafolio:
- Allows you to manually control the existence of the install target (This is more in line with other vulkan sdk adjacent code bases and super useful for projects that manually pull in Vulkan-Utility-Libraries as a source dependency where you still might want the install target but added it via cmake `add_subdirectory`)
- defaults to `VUL_IS_TOP_LEVEL` to keep the original behavior without any changes to existing client code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,11 @@ project(VUL LANGUAGES CXX)
 
 string(COMPARE EQUAL ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_SOURCE_DIR} VUL_IS_TOP_LEVEL) # Remove when min is 3.21
 
+# Control whether to install or not. By default its only on if this is the top level cmake project.
+if(NOT DEFINED VUL_ENABLE_INSTALL)
+    option(VUL_ENABLE_INSTALL "Enable install" ${VUL_IS_TOP_LEVEL})
+endif()
+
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -9,6 +9,11 @@ project(VUL LANGUAGES CXX)
 
 string(COMPARE EQUAL ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_SOURCE_DIR} VUL_IS_TOP_LEVEL) # Remove when min is 3.21
 
+# Control whether to install or not. By default its only on if this is the top level cmake project.
+if(NOT DEFINED VUL_ENABLE_INSTALL)
```
charles-lunarg, line 13: Why is an if statement put around the install code? Shouldn't making the install option be FALSE if VUL_IS_TOP_LEVEL is false be sufficient?

Put another way, we do not need to guard against the variable not existing before defining it. `option()` does not override the value if the variable already exists.
https://cmake.org/cmake/help/latest/command/option.html

mokafolio: Good Question: If I recall correctly, its mainly to make the behavior consistent with older cmake versions: https://cmake.org/cmake/help/latest/policy/CMP0077.html#policy:CMP0077

I am not married to that check, if you think its too ugly, feel free to remove it!

charles-lunarg:
Closing in favor of #273

PR #253, opened 2024-12-12
build: Add missing include headers

atetubou:
This is to fix build error when we set use_libcxx_modules=true in chromium build.


christophe-lunarg (approved):
Looks good for the layer setting library

PR #260, opened 2025-01-20
formats: Remove vkuFormatElementIsTexel

spencer-lunarg:
This `vkuFormatElementIsTexel` is a bad, misleading function

1. There is no concept of "elements" for texel in the Vulkan Spec (https://gitlab.khronos.org/vulkan/vulkan/-/issues/4109)
2. This is only used in a single VVL check for deciding if the Vertex attribute should be divided or not (which still not sure if doing correctly or not)

Basically there is no valid spec language backing this function up and if someone "really" needs it, they can manually do the check this is removing (which I am still convinced is not fully correct)

### DIFF
--- a/include/vulkan/utility/vk_format_utils.h
+++ b/include/vulkan/utility/vk_format_utils.h
@@ -270,9 +270,6 @@ inline VkExtent3D vkuFormatTexelBlockExtent(VkFormat format);
 // Returns the Compatibility Class of a VkFormat as defined by the spec
 inline enum VKU_FORMAT_COMPATIBILITY_CLASS vkuFormatCompatibilityClass(VkFormat format);
 
-// Return true if a VkFormat is 'normal', with one texel per format element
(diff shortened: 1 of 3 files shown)


include/vulkan/utility/vk_format_utils.h
```diff
@@ -270,9 +270,6 @@ inline VkExtent3D vkuFormatTexelBlockExtent(VkFormat format);
 // Returns the Compatibility Class of a VkFormat as defined by the spec
 inline enum VKU_FORMAT_COMPATIBILITY_CLASS vkuFormatCompatibilityClass(VkFormat format);
 
-// Return true if a VkFormat is 'normal', with one texel per format element
```
spencer-lunarg, line 273: `normal` here is from what I can tell just a bad comment left over from years of trying to get the original format checks correct in VVL

charles-lunarg (approved):
Happy to have nonsensical functions removed.

PR #261, opened 2025-01-21
formats: Add vkuFormatTexelBlockSize

spencer-lunarg:
`Element` is not a real spec term, it is a `Texel` inside a `Texel Block`

This adds `vkuFormatTexelBlockSize` and deprecates `vkuFormatElementSize`/`vkuFormatElementSizeWithAspect` as they should not be used.

The core issues is there are various places (like vertex inputs) where we use Formats and want the texel block size, but we don't actually need to care about if it is depth/stencil/mult-planar because  its not even allowed.... Basically people should use the utils to get the `Texel Block` size and then adjust it depending on the use case

### DIFF
--- a/include/vulkan/utility/vk_format_utils.h
+++ b/include/vulkan/utility/vk_format_utils.h
@@ -270,12 +270,19 @@ inline VkExtent3D vkuFormatTexelBlockExtent(VkFormat format);
 // Returns the Compatibility Class of a VkFormat as defined by the spec
 inline enum VKU_FORMAT_COMPATIBILITY_CLASS vkuFormatCompatibilityClass(VkFormat format);
 
+// Returns the number of bytes in a single Texel Block.
+// When dealing with a depth/stencil format, need to consider using vkuFormatStencilSize or vkuFormatDepthSize.
+// When dealing with mulit-planar formats, need to consider using vkuGetPlaneIndex.
+inline uint32_t vkuFormatTexelBlockSize(VkFormat format);
+
 // Return size, in bytes, of one element of a VkFormat
 // Format must not be a depth, stencil, or multiplane format
+// Deprecated - Use vkuFormatTexelBlockSize - there is no "element" size in the spec
 inline uint32_t vkuFormatElementSize(VkFormat format);
 
 // Return the size in bytes of one texel of a VkFormat
 // For compressed or multi-plane, this may be a fractional number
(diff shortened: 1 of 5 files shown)


include/vulkan/utility/vk_format_utils.h
```diff
@@ -2036,10 +2043,14 @@ inline VkExtent3D vkuFormatTexelBlockExtent(VkFormat format) { return vkuGetForm
...
+
+// Deprecated - Use vkuFormatTexelBlockSize
 inline uint32_t vkuFormatElementSize(VkFormat format) {
     return vkuFormatElementSizeWithAspect(format, VK_IMAGE_ASPECT_COLOR_BIT);
 }
 
+// Deprecated - Use vkuFormatTexelBlockSize
 inline uint32_t vkuFormatElementSizeWithAspect(VkFormat format, VkImageAspectFlagBits aspectMask) {
     // Depth/Stencil aspect have separate helper functions
     if (aspectMask & VK_IMAGE_ASPECT_STENCIL_BIT) {
```
charles-lunarg, line 2056: Why not just remove the functions? Deprecation comments are... unlikely to be noticed and just delay using the correct function. 

I'm fine with just yeeting the functions to speed up the process. Maybe add a comment to the new functions with "used to be called X".

spencer-lunarg: `vkuFormatElementSizeWithAspect` is used in `vkuFormatTexelSizeWithAspect` still... want to remove `vkuFormatTexelSizeWithAspect` as well, but need a replacement for that, so all of them gone soon

charles-lunarg: I'm fine with just deprecating them, but wanted to nip "forgotten deprecation notice" in the bud, IE don't forget that we wanted to remove this. Cause if we forget to remove it, then why not just keep it working long term...

spencer-lunarg: I guess asking for a "trust me bro" here, this will be gone by end of week (need working group time tomorrow to clarify a few last things before hard removing)

charles-lunarg: That's more than enough for me!

And yeah I failed to notice how one of the deprecated functions isn't just renamed, but removed 'functionality'.

charles-lunarg (approved):

PR #267, opened 2025-01-30
build: Update to header 1.4.307

mikes-lunarg:


ziga-lunarg (approved):

scripts/generators/safe_struct_generator.py
```diff
@@ -726,29 +726,25 @@ def qfi_construct(item, member):
 
             safe_name = self.convertName(struct.name)
             if struct.union:
-                if struct.name in self.union_of_pointers:
```
spencer-lunarg, line 729: @mikes-lunarg (this file code change is less worst than I was expecting TBH)

Before it gets evicted from your L2 Cache, why did this need to be removed, seems to work lol

mikes-lunarg: > @mikes-lunarg (this file code change is less worst than I was expecting TBH)

The code change is small because safe_struct unions are still kinda broken, but now `safe_VkClusterAccelerationStructureOpInputNV` isn't any more broken than the other ones

> Before it gets evicted from your L2 Cache, why did this need to be removed, seems to work lol

Because it was dead code, `self.needSafeStruct` in the loop above returns `False` for any structs in `self.union_of_pointers` so you can't even get here.

spencer-lunarg: I think a while ago, (when i did device generated commands) I proved to myself the unions in Descriptor Buffers and Device Generated Commands didn't need to do anything but save the pointer, so we didn't need any manual logic (because the pointer would always be alive and there as no reason to try having Safe Struct safe it)


... regardless, the header is all in, so time to forgot and come back and go "wtf" in a few months

PR #271, opened 2025-02-21
scripts: Use new VulkanObject from headers

spencer-lunarg:
Simple swapping of VulkanObject/BaseGenerator

cc @jeremyg-lunarg @aqnuep Here is a reference how easy it is to swap if you need to do to other repos currently using this


scripts/generate_source.py
```diff
@@ -108,6 +108,9 @@ def RunGenerators(api: str, registry: str, targetFilter: str) -> None:
     # Filter if --target is passed in
     targets = [x for x in generators.keys() if not targetFilter or x in targetFilter]
 
+    # Parse the specified registry XML into an ElementTree object
+    tree = ElementTree.parse(registry)
```
spencer-lunarg, line 112: this was unrelated, but noticed we re-ran `ElementTree` for no reason each iteration of the loop (likely was copy-and-pasted everywhere else too)

charles-lunarg (approved):

PR #275, opened 2025-02-27
Add additional Android NDK API guards

alan-baker:
* As of r28, some unguarded calls now generate `-Wunguarded-availability`
  * Change class to be guarded by ifdefs

This was caught during our internal rollout of r28


alan-baker:
cc @spencer-lunarg

spencer-lunarg (approved):

PR #276, opened 2025-03-05
formats: Refactor vkuGetFormatInfo to not use switch

charles-lunarg:
Puts the contents of a switch in a big array, then rewrites vkuGetFormatInfo() to index into it based on the formats enum value.

The logic for generating the logic of vkuGetFormatInfo is complex but boils down to finding the groups of consecutive VkFormats, where the end of a group has a value that is more than 1 away from the next format value.

~~This is a draft as the required changes to vulkan_object.py need to be made upstream (eg Vulkan-Headers). When the next header update that contains the needed changes are made, this PR can be rebased ontop of it.~~ Header has updated and this PR has been rebased, just need a review now.


arno-lunarg (approved):
LGTM!

PR #277, opened 2025-03-06
format: Fix DepthStencil to Color compatible check

spencer-lunarg:
for https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9588

I completely messed up and forgot the stencil aspect mask case for something like `VK_FORMAT_D32_SFLOAT_S8_UINT` is different formats

image


charles-lunarg (approved):
Normally changing a "stable" API is a bad thing but in this case a broken function shouldn't be used and fixing it is more important. So yeah this is an easy approve.

spencer-lunarg:
this is only for VK_KHR_maintenance8 which wasn't even part of the 1.4.301 SDK... we are breaking no one :rofl: (this was just me putting my VVL code in a shared space)

PR #282, opened 2025-03-29
formats: Handle SRGB correction

spencer-lunarg:
This is to make sure when https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/7257 is merged, this won't mess up the `vkuFormatIsSRGB` function


charles-lunarg (approved):
Had to look at the source issue to understand - but it makes sense. The Alpha channel isn't in the sRGB space, interesting! Makes sense though.

PR #293, opened 2025-05-26
enums: Add C Name helper for VkStructureType

spencer-lunarg:
For https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/10134

I need a way to print out the original C-API names from the Spec 

called it `string_VkStructureTypeCName` and open to other ideas


scripts/generators/enum_string_helper_generator.py
```diff
@@ -112,4 +112,21 @@ def generate(self):
             out.append('#endif // __cplusplus\n')
         out.extend(guard_helper.add_guard(None))
         out.append('// clang-format on')
+
+        # Special lookup for struct name of VkStructureType
+        out.append('''\n
+                   // Same thing as string_VkStructureType, but prints out the C API name
+                   static inline const char* string_VkStructureTypeCName(VkStructureType input_value) {
```
charles-lunarg, line 119: Possibly?
```suggestion
                   static inline const char* string_VkStructureTypeAPIName(VkStructureType input_value) {
```
No better names come to mind. Could try `string_pretty_VkStructureType()` but that implies all enums could be prettier, where this is the only one.

Maybe something that describes how we are going from enum to typename? `string_vulkan_type`? Because we want to map the enum back to its source type.

charles-lunarg (approved):
Leaving it to your judgement whether to rename or not. CName is kinda kludgy, but its also probably fine - there isn't any better name that comes off the top of my head. So feel free to merge at your digression.

charles-lunarg (approved):
Muuuuuch better!

PR #295, opened 2025-06-06
build: Update to header 1.4.317

mikes-lunarg:


include/vulkan/vk_enum_string_helper.h
```diff
@@ -7966,8 +8111,8 @@ static inline std::string string_VkBufferUsageFlags2(VkBufferUsageFlags2 input_v
 #endif // __cplusplus
 static inline const char* string_VkHostImageCopyFlagBits(VkHostImageCopyFlagBits input_value) {
     switch (input_value) {
-        case VK_HOST_IMAGE_COPY_MEMCPY:
-            return "VK_HOST_IMAGE_COPY_MEMCPY";
+        case VK_HOST_IMAGE_COPY_MEMCPY_BIT:
+            return "VK_HOST_IMAGE_COPY_MEMCPY_BIT";
```
charles-lunarg, line 8115: Odd but not sure if the spec just changed or there is a bug.

charles-lunarg (approved):

include/vulkan/vk_enum_string_helper.h
mikes-lunarg: Release notes indicate they cleaned up the names

> Add xml_consistency.py check for enum names extending FlagBits types but without _BIT in the enum name, fix existing violations by adding aliases, and document the naming requirement (internal issue 4276).

PR #323, opened 2025-10-01
[cmake] require >= Vulkan Headers version in VULConfig.cmake

### LINKED ISSUE #315: [cmake] explicitly "require" certain VK::Headers version
A common problem we face when vendoring OR system-including Vulkan::Headers and Vulkan::UtilityLibraries is version mismatching, which causes cryptic compile errors that are otherwise incomprehensible to the average user. A specific CMake flag, compile definition, or preprocessing check to "lock" the VulkanHeaders to a specific version or commit hash and generate a compiler (maybe CMake?) error if the versions don't match would be helpful.

crueter:
closes #315

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,7 +85,7 @@ if (VUL_ENABLE_INSTALL)
     file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/VulkanUtilityLibrariesConfig.cmake.in" [=[
         include(CMakeFindDependencyMacro)
         # NOTE: Because VulkanHeaders is a PUBLIC dependency it needs to be found prior to VulkanUtilityLibraries
-        find_dependency(VulkanHeaders REQUIRED)
+        find_dependency(VulkanHeaders @PROJECT_VERSION@ REQUIRED)
 
         @PACKAGE_INIT@
 


CMakeLists.txt
```diff
@@ -85,7 +85,7 @@ if (VUL_ENABLE_INSTALL)
     file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/VulkanUtilityLibrariesConfig.cmake.in" [=[
         include(CMakeFindDependencyMacro)
         # NOTE: Because VulkanHeaders is a PUBLIC dependency it needs to be found prior to VulkanUtilityLibraries
-        find_dependency(VulkanHeaders REQUIRED)
+        find_dependency(VulkanHeaders @PROJECT_VERSION@ REQUIRED)
```
charles-lunarg: Any particular reason to use `@PROJECT_VERSION@` instead of `${PROJECT_VERSION}`? Unless the @@ is required, I'd like to use ${} to make the file more self consistent.

crueter: Oh, I actually meant to do that but forgot about it :P

crueter: Actually, I remember why; CMake doesn't parse that string so we have to use `@@`. Otherwise it turns into:

image

PR #325, opened 2025-10-01
[ci] test windows cross comp and native runner, add linux/aarch64 test

crueter:
adds the native WoA runner to ensure native comp works, also linux/aarch64

TODO:
- maybe split windows-2022 and 2025?
- why no debug/release split or tests on windows-arm64?
- if that's resolved we can merge windows-arm64 to the main one
- why no msvc-dev-cmd on windows amd64?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,8 +19,8 @@ jobs:
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
-        config: [Debug, Release]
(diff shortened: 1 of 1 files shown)


christophe-lunarg:
Hi,

Some of these targets are handled by our internal C.I. system which is triggered when pushing a PR on this repository.

We are not planning to add more Github C.I. runner. Github C.I. is not free, once the quota is finished it gets very slow.

Thanks,
Christophe

PR #334, opened 2025-10-24
Mention `VK_Layer_*setting_name*` is deprecated and remove errant backtick

MarijnS95:
This is a followup from #332 which I would have force-pushed instead if it wasn't insta-closed.  If something is deprecated there should at least be a clear mention to discourage its use, and I didn't want to loose out on the formatting fixes either...

---

While reading https://vulkan.lunarg.com/doc/view/latest/windows/layer_configuration.html I was surprised to not see the `VK_LAYER_*setting_name*` that was mentioned at https://vulkan.lunarg.com/doc/view/latest/windows/synchronization_usage.html for `VK_LAYER_VALIDATE_SYNC`; as it turns out this front-and-center settings environment variable is actually deprecated.

This prefix variant is also mentioned for some but not all layer settings at: https://vulkan.lunarg.com/doc/view/latest/windows/khronos_validation_layer.html Which should all receive some sort of `(deprecated)` suffix to make this clear.

Because of the missing backtick the `*` used to make text italics show up verbatim, and by default markdown reflows consecutive lines as a single paragraph so these have been moved into a list to show on separate lines.

image

### DIFF
--- a/docs/layer_configuration.md
+++ b/docs/layer_configuration.md
@@ -334,9 +334,11 @@ Some settings from the settings file can also be set using environment variables
 
 The environment variable names for the layer settings have multiple variants that follows the format:
 
-   VK_`<`*`LayerVendor`*`>_`<`*`LayerName`*`>_<`*`setting_name`*`>` which take precedent over:
-   VK_`<`*`LayerName`*`>_<`*`setting_name`*`>` which take precedent over:
-   VK_`<`*`setting_name`*`>`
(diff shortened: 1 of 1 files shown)


spencer-lunarg:
cc @charles-lunarg @christophe-lunarg

charles-lunarg:
I will need the CLA license agreed to before I can merge this.

christophe-lunarg (approved):

PR #371, opened 2026-04-07
safe: Use Effective extended flags

spencer-lunarg:
for future extending flags extensions, I think it would be better to just have the "effective" flag value in safe struct

From validation, I find the ONLY 2 reasons we want the "original" value is

1. To deduce the `Location` (which we can still do checking if pNext has the struct on an error)
2. when we are creating the object (so `vkCreateBuffer`) but at that point, we have not made a safe struct copy yet

The whole "idea" of safe struct was to make our own internal copy and I can't think (so yell at me with it if you can think) why Safe Struct should **not** have the effective flag value

### DIFF
--- a/include/vulkan/utility/vk_safe_struct.hpp
+++ b/include/vulkan/utility/vk_safe_struct.hpp
@@ -301,7 +301,8 @@ struct safe_VkBufferCreateInfo {
     const void* pNext{};
     VkBufferCreateFlags flags;
     VkDeviceSize size;
-    VkBufferUsageFlags usage;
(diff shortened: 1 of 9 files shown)


ziga-lunarg (approved):

spencer-lunarg:
... so this failed, going `create_info.ptr()` you get a `VkBufferCreateInfo` that is not correct because the usage field went from 32 bits to 64 bits messing up the struct

decided this isn't the place to do it... Quake 2 level doom as I said

PR #374, opened 2026-04-14
Revert "safe: Remove unused includes"

charles-lunarg:
Reverts KhronosGroup/Vulkan-Utility-Libraries#373

### DIFF
--- a/include/vulkan/utility/vk_safe_struct.hpp
+++ b/include/vulkan/utility/vk_safe_struct.hpp
@@ -16,6 +16,8 @@
 
 #pragma once
 #include <vulkan/vulkan.h>
+#include <algorithm>
+
 #include <vulkan/utility/vk_safe_struct_utils.hpp>
 
 namespace vku {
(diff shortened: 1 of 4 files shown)


ziga-lunarg (approved):

spencer-lunarg:
..... this is a dumb (lazy) revert, we should have just added the `<vector>` in VVL

include/vulkan/utility/vk_safe_struct.hpp
dneto0, line 27: The original commit just landed in our laps.

When looking at it, I would have added `#include <vector>` in this file, following include-what-you-use.

dneto0 (commented):
thanks for the revert. I expect it will fix breakage for us.

include/vulkan/utility/vk_safe_struct.hpp
spencer-lunarg: @dneto0 so we ended up just removing the logic that even needed the vector 

Hopefully everyone now is fixed, if things are still not, ping me offline

PR #377, opened 2026-04-24
safe: Handle ppUsageCounts with code gen

spencer-lunarg:
Automate the `ppUsageCounts` (need for the KHR Opacity Micromap as well, so getting it in now) 

removes manual safe struct code that no one was going to maintain


spencer-lunarg:
used AI (handed it the diff) to confirm the logic is the same from Manual to Generated

ziga-lunarg (approved):

PR #378, opened 2026-05-01
Update to header v1.4.350

mikes-lunarg:


src/vulkan/vk_safe_struct_core.cpp
```diff
@@ -580,6 +601,12 @@ void safe_VkDeviceCreateInfo::initialize(const VkDeviceCreateInfo* in_struct, [[
             pQueueCreateInfos[i].initialize(&in_struct->pQueueCreateInfos[i]);
         }
     }
+    if (enabledLayerCount && in_struct->ppEnabledLayerNames) {
+        ppEnabledLayerNames = new char[enabledLayerCount];
+        for (uint32_t i = 0; i < enabledLayerCount; ++i) {
+            ppEnabledLayerNames[i] = in_struct->ppEnabledLayerNames[i];
+        }
+    }
```
charles-lunarg: This is quite a bit different than the original code
```c++
    char** tmp_ppEnabledLayerNames = new char*[in_struct->enabledLayerCount];
    for (uint32_t i = 0; i < enabledLayerCount; ++i) {
        tmp_ppEnabledLayerNames[i] = SafeStringCopy(in_struct->ppEnabledLayerNames[i]);
    }
    ppEnabledLayerNames = tmp_ppEnabledLayerNames;
```

I believe reverting in full or part https://github.com/KhronosGroup/Vulkan-Utility-Libraries/commit/9b9b0a04b36344a85705fe30f52e6f7aa59a9001 will get us a lot of the way there.

mikes-lunarg: Hooray, we can remove the hack we put in for 347

charles-lunarg (approved):

PR #380, opened 2026-05-04
Revert "scripts: NULL empty strings array"

spencer-lunarg:
This reverts commit a96bd4933c6adfc21329d15e9999791a0153bbf3.

see https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/12185#pullrequestreview-4219780481

@arno-lunarg which test was giving you a self validation error... that test might be just wrong now, **I will not merge this** yet, but want to have this ready as I don't agree with the previous change

(sorry, I'm already pissed the working group decided to mess around with `ppEnabledLayerNames` after it was not broken for 10 years and now we are all having to play stupid code gen dances around a fake problem)

### DIFF
--- a/scripts/generators/safe_struct_generator.py
+++ b/scripts/generators/safe_struct_generator.py
@@ -575,16 +575,11 @@ void FreePnextChain(const void *pNext) {
                                 # Create deep copies of strings
                                 if member.length:
                                     copy_strings += f'''
-                                        if (in_struct->{member.length} > 0) {{ 
-                                            char **tmp_{member.name} = new char *[in_struct->{member.length}];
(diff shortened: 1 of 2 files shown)


arno-lunarg:
Pretty much any test should trigger that issue with self validation, as long as the `VkDeviceCreateInfo` the test uses has `0` and `NULL` ppEnabledLayerNames and corresponding count. Self validation will intercept the safe struct made from it and thus trigger validation error

arno-lunarg:
Also this fix aside, we should not try to allocate memory for empty arrays, so I would vote for keeping those changes anyway

PR #382, opened 2026-05-08
Update to header v1.4.351

mikes-lunarg:


spencer-lunarg (approved):
4 vendors extensions and nothing broke... wow

PR #388, opened 2026-06-11
Revert "utility: Use insert_or_assign in concurrent unordered map"

charles-lunarg:
Reverts KhronosGroup/Vulkan-Utility-Libraries#387

Was found to cause build errors in chrome.

### DIFF
--- a/include/vulkan/utility/vk_concurrent_unordered_map.hpp
+++ b/include/vulkan/utility/vk_concurrent_unordered_map.hpp
@@ -62,7 +62,7 @@ class unordered_map {
     void insert_or_assign(const Key &key, Args &&...args) {
         uint32_t h = ConcurrentMapHashObject(key);
         WriteLockGuard lock(locks[h].lock);
-        maps[h].insert_or_assign(key, {std::forward<Args>(args)...});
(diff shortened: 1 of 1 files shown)


include/vulkan/utility/vk_concurrent_unordered_map.hpp
```diff
@@ -62,7 +62,7 @@ class unordered_map {
     void insert_or_assign(const Key &key, Args &&...args) {
         uint32_t h = ConcurrentMapHashObject(key);
         WriteLockGuard lock(locks[h].lock);
-        maps[h].insert_or_assign(key, {std::forward<Args>(args)...});
```
artem-lunarg, line 65: I think this failed because {std::forward<Args>(args)...} does not define a type, so adding T before probably (with enough luck) will help. And version with [] - it returns reference to existing/newly inserted element, so the type is known. Btw, original version from Gemini had T (that Spencer provied in internal chat)

artem-lunarg (approved):
Spencer is AFI. I can approve in case he is not back soon. Then we can decide if we want to fix this.

PR #399, opened 2026-08-03
cmake: support shared utility libraries

wolfv:
Allow `VulkanLayerSettings` and `VulkanSafeStruct` to follow the standard CMake `BUILD_SHARED_LIBS` option instead of always forcing static libraries.

`WINDOWS_EXPORT_ALL_SYMBOLS` supplies import libraries for the existing C++ APIs on Windows. The default remains unchanged (`BUILD_SHARED_LIBS=OFF`).

This is already exercised on Linux, macOS, and Windows by the conda-forge Vulkan Utility Libraries package: https://github.com/conda-forge/vulkan-utility-libraries-feedstock

### DIFF
--- a/src/layer/CMakeLists.txt
+++ b/src/layer/CMakeLists.txt
@@ -5,7 +5,13 @@
 # SPDX-License-Identifier: Apache-2.0
 set(CMAKE_FOLDER "${CMAKE_FOLDER}/VulkanLayerSettings")
 
-add_library(VulkanLayerSettings STATIC)
+add_library(VulkanLayerSettings)
+if(BUILD_SHARED_LIBS)
+    set_target_properties(VulkanLayerSettings PROPERTIES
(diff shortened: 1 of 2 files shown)


wolfv:
I extended the change to override the project-wide hidden visibility for shared targets. A downstream macOS consumer exposed that merely selecting `SHARED` is insufficient: the dylib otherwise contains no usable public symbols. Shared builds now use default C++ symbol visibility (and CMake-generated Windows exports), while static builds retain the existing visibility settings.

charles-lunarg (approved):
I don't see any particular reason why someone would want shared versions of these libraries, but that doesn't mean the option can't exist for those who do have a reason.

Thank you for making sure the symbols are visible, I probably wouldn't have checked that myself. (and then been annoyed with myself when a subsequent issue came in showing that I should have checked)

PR #400, opened 2026-08-05
build: Avoid AppleClang using the wrong headers

charles-lunarg:
Because of AppleClang always adding /usr/local/include and CMake using -isystem by default in 3.25+, builds on MacOS could use system installed headers by mistake. Because this repo has code generated against a specific version of the Vulkan Headers, if the system installed headers are older (which is usually the case for developers of this repo) then the build fails.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,11 @@ add_subdirectory(scripts)
 
 find_package(VulkanHeaders ${PROJECT_VERSION} CONFIG)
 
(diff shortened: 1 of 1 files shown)


mikes-lunarg (approved):
LGTM!

jrprice:
I don't know if we are holding the CMake configuration wrong, but this has caused an error in our downstream project (Dawn):
```
CMake Error at third_party/vulkan-utility-libraries/src/CMakeLists.txt:31 (set_target_properties):
  set_target_properties can not be used on an ALIAS target.
```

PR #402, opened 2026-08-08
build: Check if Vulkan-Headers is an ALIAS target

charles-lunarg:
When projects make Vulkan-headers available using add_subdirectory, the Vulkan::Headers target is an ALIAS, not an IMPORTED target. This breaks the AppleClang workaround as CMake cannot modify the properties of ALIAS targets. Because we don't need the workaround when users add Vulkan-Headers using add_subdirectory, just check that it isn't an ALIAS target before calling set_target_properties.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,7 +28,10 @@ find_package(VulkanHeaders ${PROJECT_VERSION} CONFIG)
 
 # Workaround AppleClang using the wrong headers when there are headers in /usr/local/include
 if (CMAKE_CXX_COMPILER_ID STREQUAL AppleClang AND TARGET Vulkan::Headers AND CMAKE_VERSION VERSION_GREATER_EQUAL 3.25)
-    set_target_properties(Vulkan::Headers PROPERTIES SYSTEM OFF)
+    get_target_property(vulkan_headers_aliased Vulkan::Headers ALIASED_TARGET)
+    if (NOT vulkan_headers_aliased)
+        set_target_properties(Vulkan::Headers PROPERTIES SYSTEM OFF)
+    endif()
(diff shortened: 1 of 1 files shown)


charles-lunarg:
@jrprice This should hopefully solve the issue in Dawn. But because I rolled the AppleClang workaround to a bunch of other vulkan repo's (which plays afoul of the set_target_properties issue), does Dawn use other Vulkan projects? That way I know which other projects to go and fix.

charles-lunarg:
Linking the comment to ensure future visibility for people looking through the issues
https://github.com/KhronosGroup/Vulkan-Utility-Libraries/pull/400#issuecomment-5221463133

ziga-lunarg (approved):

jrprice:
Thanks for fixing this!

> does Dawn use other Vulkan projects

Yes, we use Vulkan-Loader, Vulkan-Utility-Libraries, and Vulkan-Validation-Layers.

charles-lunarg:
> Thanks for fixing this!

No problem! 

Though, did you test the fix? I'm happy to do the same change in Vulkan-Loader and Vulkan-ValidationLayers, but curious if you are seeing the same problems there. (you might not see errors there until Vulkan-Utility-Libraries is fixed, so I'll get on the PR's regardless)

charles-lunarg:
Both Vulkan-Loader and Vulkan-ValidationLayers have been updated, @jrprice

When you get a chance, it'd be great if you could confirm Dawn builds without issue.

jrprice:
> When you get a chance, it'd be great if you could confirm Dawn builds without issue.

Confirmed that Dawn is now building for me again on macOS without issue, thanks!

PR #408, opened 2026-09-21
Allow generating a Vulkan SC scoped variant

jeremyg-lunarg:
-  scripts: Allow generating a Vulkan SC scoped variant

    Accept --api vulkansc, and add --no-merge to generate for that API alone
    rather than merged with base Vulkan.

    Two parts of this are tricky:

    - vk.xml tags VK_VERSION_1_3 and 1_4 for vulkansc, even though VKSC_VERSION_1_0
      depends only on VK_VERSION_1_2. GetApiVersions walks `depends` to find
      the versions the variant really uses.

    - The safe struct sources use extension types unconditionally. Guard the
      ray tracing, acceleration structure, VK_EXT_descriptor_heap and QCOM
      image processing code on their extensions. safe_VkGraphicsPipelineCreateInfo
      is core, so its VK_EXT_graphics_pipeline_library lookup instead goes
      through a helper returning false when the extension is absent.

- scripts: Limit format utils to formats the API enables

    The <formats> catalog is not API-scoped, the VkFormat enum is.
    Only emit the formats that are valid for the current api.

- scripts: Close trailing platform guard in string_VkStructureName

    The loop never calls add_guard(None), leaving its last #ifdef open. This
    is currently harmless because the last sType emitted has no guard.


jeremyg-lunarg:
Sigh. I forgot there was alreadyVulkanSC-Utility-Libraries

PR #409, opened 2026-09-21
Fix clang-format problems

jeremyg-lunarg:
- style: Set DerivePointerAlignment explicitly
-  ci: Pin the clang-format version

These make sure we get consistent clang-format results on machines running different versions of the tool.

- ci: Pin the clang-format version

This fixes a globbing bug in the ci script which could cause some file to skip checking.


spencer-lunarg (commented):
Personally I rather do what VVL does and use clang-format 22 (as its the newest and easy to grab for windows)

https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/11921/files

then if you are in Ubuntu 24.04 (like I am) I added this

https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/11941

so we don't waste time thrashing the clang-format

if you add this change, it is annoying to make people on Fedora or Windows "downgrade" to match clang-format 18

jeremyg-lunarg:
I'm fine with using a newer version of clang-format.  But why not just make the one line change in [5468aea](https://github.com/KhronosGroup/Vulkan-Utility-Libraries/pull/409/commits/5468aeab6a3d5a4aa6cc0d18573d701c7bb89e68) which stops the format churn?

spencer-lunarg:
> But why not just make the one line change in which stops the format churn?

... I am like 99% sure I tried ALL the stupid `.clang-format`  settings and non worked... can you confirm that if you have that on and run clang-format 18 and something new like 22 or 23 that it still doesn't trash back and forth

jeremyg-lunarg:
> > But why not just make the one line change in which stops the format churn?
> 
> ... I am like 99% sure I tried ALL the stupid `.clang-format` settings and non worked... can you confirm that if you have that on and run clang-format 18 and something new like 22 or 23 that it still doesn't trash back and forth

yes with the setting I get ths same results for all versions from 18 to 23. Without the setting I get ~1300 lines changed for 21 and up. Worst file is  include/vulkan/utility/vk_sparse_range_map.hpp

CDL requires an additional setting `AllowShortBlocksOnASingleLine: Empty` to avoid this churn:

```
 -        { os << YAML::Comment("StdVideoH264SequenceParameterSet"); }
+        {
+            os << YAML::Comment("StdVideoH264SequenceParameterSet");
+        }
```

Maybe there's other stuff in VVL that isn't covered.

spencer-lunarg:
for me it was the `const Object* x` vs `const Object *x` that was causing me constant churn

spencer-lunarg:
.... if you confirmed it works here, cool, I still think we should use newer clang-format in CI and if the `.clang-format` is going to work here, I'm happy with this then

spencer-lunarg (approved):

